PR #79, opened 2023-10-27
Elixir source code improvements (except tests)

Eiji7:
Hi, I made a code style rewrite for `Elixir` source (except tests). :spiral_notepad: 

Here is a list of edit reasons :exploding_head: 
- [**updated**] credo config
- [**fixed**] Credo.Check.Refactor.ABCSize by excluding tap and then functions (similar to default's ecto macros exclude)
- [**updated**] environment, Elixir and JS dependencies
- [**refactor**] simplyfi reducers
- [**refactor**] simplyfi code using reducers
- [**refactor**] use Kernel.tap/2 and Kernel.then/2 when applicable (piepelines are much easier to split, debug and edit - this also causes less temporary variables)
- [**refactor**] use UNIX "find … | sort | uniq" as a faster and simpler solution
- [**refactor**] rewritten path concatenation using Path.join/1 and Path.join/2 calls
- [**refactor**] rewritten build_dir by removing root_path call in favor of safer :code.lib_dir/2 call
- [**refactor**] simplify code by moving shared parts to private functions
- [**refactor**] replaced usage of operators called with Kernel in pipeline in favor of Kernel.then/1 macro
- [**fixed**] follow credo's safer suggestions as long as they don't cause test fail (review needed?)
- [**refactor**] rewritten usage of case, cond, and if control structures in favor of Kernel.tap/2 and Kernel.then/2 macros
- [**refactor**] rewritten code using Kernel.inspect/1 when applicable
- [**refactor**] rewritten code using cleaner (and faster in some cases) pattern matching when applicable
- [**added**] Hologram.Commons.StringUtils.append/2
- [**added**] Hologram.Commons.StringUtils.unwrap/3
- [**fixed**] rewritten npm calls to hide unrelevant audit/fund/progress information
- [**refactor**] replaced nested Enum.each/2 calls in favor of for comprehension
- [**fixed**] updated encoded urls in comments
- [**refactor**] rewritten encode_as_string/1 code in favor of more clear and safe StringUtils.unwrap/3 call
- [**refactor**] some functions has been splitted into smaller ones to increase readability and fix credo warnings
- [**refactor**] rewritten cryptic integer code points in favor of unicode code point syntax
- [**refactor**] rewritten single-letter module alias used in function specifications
- [**refactor**] expanded maybe_add_default_bitstring_type_modifier/2 into more detailed logic
- [**refactor**] move big multiline strings outside of function definition if possible
- [**cleanup**] move credo disable common on IoPuts check to the start of each mix task file if applicable
- [**cleanup**] updated test for recusively listing files since now this function returns empty list instead of raising error
- [**cleanup**] ensure npm, credo, dialyzer and test --trace does not produces and warning or error

Everyone have their own preferences, but I hope you see the power of pipelines powered by `Kernel.tap/1` and `Kernel.then/1`. In other cases I tried to match your style (like extending `StringUtils` common helpers). If you like my input I can think about making a similar improvements for `tests` and/or maybe by adding some features. :see_no_evil: 

I saw you are not aware of `?` syntax. For more information see: [Integers in other bases and Unicode code points](https://hexdocs.pm/elixir/syntax-reference.html#integers-in-other-bases-and-unicode-code-points) :link: 

From my side of view PR is ready to merge as there are no conflicts and all checks passes. Feel free to ask any question. I'm leaving `Allow edits by maintainer` option enabled. :+1: 

Keep calm and … **love `Elixir`** (and it's community). :heart: 

All the best! Hope for hearing release news from you soon! :rocket: 

Best regards,
Tomasz Sulima
Ramurix Software SRL


bartblast:
Hey, thanks a bunch for diving into this — it's clear you put in some serious effort!

I'd prefer not to merge it in right now, unfortunately, for a couple of reasons:

1)
A big chunk of the updates deals with code style, which we haven’t discussed. I was planning to tackle this post-0.2.0 release when I start engaging collaborators. For instance, I'm not sold on using Kernel.tap/2 and Kernel.then/2 as you did in the rewrite for every situation. I see their place in specific cases, but I find longer code or temporary variables often make things more readable and communicate intent better. Noticed a bunch of updates aimed at fitting these functions into pipelines.

2)
This pull request is MASSIVE! ;) I'm thinking breaking it down into separate ones, each tackling a specific issue might be easier to process. That way, we could discuss if it's necessary before diving into the code. Plus, pull requests are all-or-nothing — I can either accept or decline the whole thing.

3)

Until the 0.2.0 release (the playable version), I'm trying to keep things low on the project management side. Wanna keep the focus 100% on coding. However, small bug fixes, doc tweaks, or no-brainer / uncontroversial updates are more than welcome.

---

Most importantly, I don't want you thinking I'm not appreciative of your work or feedback. Quite the opposite! I'm looking for contributors and even building a list of tasks for them as I find something that fits. Just trying to keep things streamlined for now.

By the way, I'm genuinely thrilled that people recognize the value in this project — enough to invest so much effort into making it even better.

Thanks a ton again! 💜

PR #81, opened 2023-11-07
Split CI Lib and Feature tests

gerbal:
This PR splits the CI jobs in two, allowing the library test suite and the feature tests suit to run in parallel

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ permissions:
 jobs:
   lib_test:
     runs-on: ubuntu-latest
-    name: Test on OTP ${{ matrix.otp }} / Elixir ${{ matrix.elixir }}
+    name: Lib Test on OTP ${{ matrix.otp }} / Elixir ${{ matrix.elixir }}
     strategy:
       # Specify the OTP and Elixir versions to use when building
       # and running the workflow steps.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,9 +17,9 @@ permissions:
   contents: read
 
 jobs:
-  test:
+  lib_test:
     runs-on: ubuntu-latest
-    name: Test on OTP ${{ matrix.otp }} / Elixir ${{ matrix.elixir }}
+    name: Lib Test on OTP ${{ matrix.otp }} / Elixir ${{ matrix.elixir }}
```
bartblast: Can you make it plural? I see that I used singular initially, but after splitting it it seems that it's better to use singular:
lib_test -> lib_tests, Lib Test -> Lib Tests, feature_test -> feature_tests, etc.

.github/workflows/ci.yaml
```diff
@@ -198,27 +229,27 @@ jobs:
     # Step: Download feature tests app Elixir dependencies.
     # If unchanged, uses the cached version.
     - name: Download feature tests app Elixir deps
-      run: cd test/features && mix deps.get
+      run:  mix deps.get
```
bartblast: Can you remove redundant spacing after "run:  " (this happens on all the updated lines)

bartblast (changes requested):
Thanks @gerbal, just a few minor change requests.

bartblast:
Thank you! 💜

PR #157, opened 2025-01-09
Fix CI by removing unused dependency

PragTob:
Heyo, couldn't help but notice the CI was failing due to an unused dependency so thought I'd go and fix that really quickly.

https://github.com/bartblast/hologram/actions/runs/12683176110/job/35349863480

Good luck and thanks for your efforts!


bartblast:
Thanks for help Tobi!

PR #158, opened 2025-01-10
Remove missed libgraph dep from mix.lock

kevinam99:
Hey! This is Kevin from the Elixir Slack.

Just saw that the CLI was failing again so I wanted to check what's causing it. I found that libgraph is an unused dependency in an additional mix.lock file which was missed so I went ahead and removed it from there. 

I hope this helps!


bartblast:
Oh yeah, that seems to be the same issue as here: https://github.com/bartblast/hologram/pull/157 I had to vendor libgraph, since there hasn't been a hex release for a long time and Hologram used a GitHub ref. Thanks for the help, Kevin! :)

kevinam99:
Oh, makes sense. I saw that libgraph is vendored but didn't quite understand why.

PR #164, opened 2025-01-18
Fix reraise call

Eiji7:
The `reraisee/2` function expects **stacktrace** as a second argument.
In order to pass error attributes we have to use `reraise/3` with **stacktrace** as a 3rd argument.

See documentation of  [reraise/3](https://hexdocs.pm/elixir/Kernel.html#reraise/3) for more information.

### DIFF
--- a/lib/hologram/page.ex
+++ b/lib/hologram/page.ex
@@ -113,9 +113,10 @@ defmodule Hologram.Page do
     String.to_existing_atom(value)
   rescue
     ArgumentError ->
-      reraise Hologram.ParamError,
-        message:
-          ~s/can't cast param "#{name}" with value #{KernelUtils.inspect(value)} to atom, because it's not an already existing atom/
+      message =
+        ~s/can't cast param "#{name}" with value #{KernelUtils.inspect(value)} to atom, because it's not an already existing atom/
+
+      reraise Hologram.ParamError, [message: message], __STACKTRACE__
   end
 
   defp cast_param(:atom, value, name) do


bartblast:
Thanks @Eiji7, would you mind checking if this problems affects the client-side as well? Probable files to look at: interpreter.mjs, renderer.mjs and possibly test JS helpers such as: defineElixirHologramRouterHelpersModule?

Eiji7:
Oh right, I did not looked at client-side code.

**Noe**: As wrote on forum I'm currently a bit busy recently. I will try to reply asap, but please expect possible delays between reply time and changes in code. I always keep `Allow edits by maintainers` enabled just in case you need to make some changes asap.

Eiji7:
@bartblast I just made a `grep` looking for `reraise` word and I found that there is one more call in `lib/hologram/compiler/encoder.ex`, but this one was correct. Perhaps you have not yet implemented `reraise` in `JavaScript`?

```bash
$ grep -ir reraise {assets,lib,test}
lib/hologram/compiler/encoder.ex:      reraise RuntimeError, [message: message], __STACKTRACE__
lib/hologram/page.ex:      reraise Hologram.ParamError, [message: message], __STACKTRACE__
```

bartblast:
Allright, I took a deeper look and found that these changes only affect the backend (param processing happens server-side). The encoder is strictly compiler-related. I've added the missing test case here: https://github.com/bartblast/hologram/commit/cc264483b287ed453c44d1d4421f8ce86c7caf53
I also cleaned up some outdated unit tests and narrowed down the params type spec here: https://github.com/bartblast/hologram/commit/774c14a3b42819813f0ddfb57955ec15296e55ed
Thanks for your help @Eiji7!

PR #166, opened 2025-01-22
Update project version reference in project's docs configuration

Eiji7:
I made it easier by storing it in module attribute, so both `:version` and `:source_ref` in `:docs` uses same version.

If possible please move `v0.2.0` after merging this commit and update `hexdocs`, so the links to source code would use specified tag instead of `master`. However we can make it `v0.2.1` or so if needed to push docs.

Also looks like you have set `0.3.0` version in project configuration. I would recommend to change it right before creating new tag and pushing new release to `hex`.

**Edit**: I just found that you use `0.2.0` as latest release and `0.3.0` as latest tag and that's a bit confusing. Please let me know if I should change the module attribute value to `"0.3.0"` or we still want to work on `0.2.0`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -2,7 +2,7 @@
 defmodule Hologram.MixProject do
   use Mix.Project
 
-  @version "0.2.0"
+  @version "0.3.0"
 
   defp aliases do
     [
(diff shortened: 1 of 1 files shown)


bartblast:
Hey @Eiji7, thanks. I added v0.3.0 tag, because I'm releasing 0.3.0 right now, can you change the version to 0.3.0?

Eiji7:
@bartblast done

PR #226, opened 2025-08-02
Update change_event.mjs

absowoot:
Return bool for checkbox, otherwise checkboxes will always have a value of "on".


bartblast:
Thank you @absowoot! :)

PR #249, opened 2025-09-01
[WIP] Sets module implementation

Lucassifoni:
Hello Bart,

I have a first draft that works when used from Elixir through MapSet (at least in terms of new/put). Before going through the full :sets module implementation and writing all the required tests, I wanted to check if what I understood of Hologram's conventions seemed right to you.

Best,
Lucas

### DIFF
--- a/assets/js/erlang/sets.mjs
+++ b/assets/js/erlang/sets.mjs
@@ -1,0 +1,122 @@
+"use strict";
+
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+import Erlang_Maps from "./maps.mjs";
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+// :sets v2 functions work with maps internally
+// The set is represented as a map where keys are the elements and values are empty lists []
+
+const Erlang_Sets = {
+  // Start new/0
+  "new/0": () => {
+    // Return a map (version 2 representation)
+    return Type.map([]);
+  },
+  // End new/0
+  // Deps: []
+
+  // Start new/1
+  "new/1": (opts) => {
+    if (!Type.isList(opts)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":sets.new/1", [opts]),
+      );
+    }
+
+    return Type.map([]);
+  },
+  // End new/1
+  // Deps: []
+
+  // Start from_list/2
+  "from_list/2": (list, opts) => {
+    if (!Type.isList(list)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":sets.from_list/2", [
+          list,
+          opts,
+        ]),
+      );
+    }
+
+    if (!Type.isList(opts)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":sets.from_list/2", [
+          list,
+          opts,
+        ]),
+      );
+    }
+
+    let set = Type.map([]);
+    const emptyList = Type.list([]);
+
+    for (const element of list.data) {
+      if (!Type.isTrue(Erlang_Maps["is_key/2"](element, set))) {
+        set = Erlang_Maps["put/3"](element, emptyList, set);
+      }
+    }
+
+    return set;
+  },
+  // End from_list/2
+  // Deps: [:maps.is_key/2, :maps.put/3]
+
+  // Start add_element/2
+  "add_element/2": (element, set) => {
+    if (!Type.isMap(set)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":sets.add_element/2", [
+          element,
+          set,
+        ]),
+      );
+    }
+    if (Type.isTrue(Erlang_Maps["is_key/2"](element, set))) {
+      return set;
+    }
+
+    return Erlang_Maps["put/3"](element, Type.list([]), set);
+  },
+  // End add_element/2
+  // Deps: [:maps.is_key/2, :maps.put/3]
+
+  // Start del_element/2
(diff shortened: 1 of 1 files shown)


bartblast:
Hey Lucas!

Here are a few suggestions:

1. `new/0` - not sure we should support it for now, since it's not used and works with `:sets` version 1 only
2. `new/1` - if we're supporting only version 2, you need to validate the allowed options. See how it's done for `:unicode.characters_to_binary/3` (which currently allows only `"utf8"`). Generally if some option is not supported, always validate and raise a clear error message that it's not supported yet in Hologram.
3. `from_list/2` - same remark about opts validation. Also maybe we can use `:maps.from_list/2` if only sets version 2 is used? You'd basically need to build something like this for the arg: `[{:elem1, []}, {:elem2, []}, ...]` - though this is just my idea from top of my head at first sight, I didn't verify this.

Generally when working with collections there may be a big performance penalty when building new objects every time for each loop pass, so if you create a new var inside the ported function don't shy away from mutating it and building the final boxed value in place. For that you need to know the internal boxed values structure (see `Type` class functions and for example `Type.encodeMapKey()`).

As for conventions, all looks good - just one minor correction: order imports and functions in the module alphabetically :)

Great work overall!

Lucassifoni:
Hello Bart, thanks for your answer - the part about internal mutation is one of the things I had in mind.

I would also be keen about not supporting `:sets` version 1.

For background trivia, the website I'm looking to port is https://audioskop.pierredamienhuyghe.fr/map . (WIP)
Hologram would make a lot of sense for the whole user-facing part. For the map operations, at a glance I'm using almost all of Elixir's MapSet API.

I'll continue further with proper mirroring of the :sets API and options.

Lucassifoni:
Hi @bartblast, just to let you know that I am still on that – just had a very busy period with contract work. I am getting back on it.

bartblast:
Hey @Lucassifoni, great, thanks for letting me know!

Lucassifoni:
Hi @bartblast,

Re: :sets version 1 (tuple buckets representation) or :sets version 2 (map with empty list as values) : 

Elixir users with OTP 27 and below will default to the version 1 under the hood, while OTP 28 starts using version 2. On the other hand, Elixir's MapSet module had a bespoke implementation of sets (actually more or less identical to :sets version 2) until Elixir 1.15, then switched to using erlang's :sets module.

If I do some structural assertions on the data shapes sets operation result in, in the javascript sets tests, I cannot then fully mirror them elixir-side for the consistency tests as the data shape will vary among elixir users. 

I think the intersection of Hologram users and people who would rely on the internal representation of sets despite it being actively warned about in OTP docs is very small, so I think we could support both versions in the public API of the JS implementation, while having a single one under the hood (the more recent map one).

For the elixir consistency tests, we would need to do higher-level assertions than the JS ones, while keeping the tests themselves coherent. Is that okay ?

bartblast:
Hey Lucas!

I think we should not support version 1 at all. As you noticed, the internal representation is predominantly internal, and where version is given as an option, we can raise a clear exception that version 1 isn't supported.

Managing different OTP implementations is something we could do in the future, but it's definitely too early. For now, let's focus on as much coverage as possible for typical use cases.

If some functions implicitly use different versions depending on OTP/Elixir version (and that makese tests harder), you can execute different assertions using `System.otp_release/0 `or something like `Version.match?(System.version(), ">= 1.17.0")` for Elixir versions. Here's an example: https://github.com/bartblast/hologram/blob/a1a1dba83c635f7a35efb46c6b4ba3d9919c4664/test/elixir/hologram/compiler/transformer_test.exs#L1895

One more thing - could you split the functions into multiple PRs? This will be much more manageable for me to review, and I'll be able to help you as you work on each function. Thanks! :)

Lucassifoni:
Hello Bart,If having a few OTP-detecting utilities in tests is okay for you then this solves this problem entirely. I will test under otp 24 to 28.I will split in 1 PR per function, we can close this one then.Best,LucasLe 21 nov. 2025 à 19:54, Bart Blast ***@***.***> a écrit :﻿bartblast left a comment (bartblast/hologram#249)
Hey Lucas!
I think we should not support version 1 at all. As you noticed, the internal representation is predominantly internal, and where version is given as an option, we can raise a clear exception that version 1 isn't supported.
Managing different OTP implementations is something we could do in the future, but it's definitely too early. For now, let's focus on as much coverage as possible for typical use cases.
If some functions implicitly use different versions depending on OTP/Elixir version (and that makese tests harder), you can execute different assertions using System.otp_release/0 or something like Version.match?(System.version(), ">= 1.17.0") for Elixir versions. Here's an example: https://github.com/bartblast/hologram/blob/a1a1dba83c635f7a35efb46c6b4ba3d9919c4664/test/elixir/hologram/compiler/transformer_test.exs#L1895
One more thing - could you split the functions into multiple PRs? This will be much more manageable for me to review, and I'll be able to help you as you work on each function. Thanks! :)

bartblast:
> Hello Bart,If having a few OTP-detecting utilities in tests is okay for you then this solves this problem entirely. I will test under otp 24 to 28.I will split in 1 PR per function, we can close this one then.Best,Lucas

Great, thanks!

Lucassifoni:
The last elixir release available for OTP 23 is 1.14.4, already quite old. Sets V2 were introduced in OTP 24. I tried compiling Hologram under elixir 1.14.4-otp-23 ~~and it failed on some dependencies (yaml_elixir notably).~~

Can I safely disregard OTP versions under 24, before the introduction of sets V2 ?

Lucassifoni:
Edit : the ~HOLO sigil requires elixir 1.15 since it introduced multi-letter sigils. Since there is no OTP release under 24 that supports it, I will disregard OTP < 24 but leave a note in the test files.

bartblast:
Yeah, we support Elixir 1.15+ /  OTP 24+ / Node 20+, see: https://github.com/bartblast/hologram/blob/86fcb19d25815123ef8f179c57485ddaaf072213/.github/workflows/ci.yml#L29

Just added this info to the Prerequisites section on the website Installation page: https://hologram.page/docs/installation

bartblast:
@Lucassifoni, Can you comment anything in the following issue (e.g. "Got it."), otherwise I can't assign you properly:
* https://github.com/bartblast/hologram/issues/295
* https://github.com/bartblast/hologram/issues/296
* https://github.com/bartblast/hologram/issues/297

PR #250, opened 2025-09-06
Introduce template formatter

mwmiller:
This handles, in some way, everything which the parser does.  We could probably use some tests, documentation and improved configurability.


mwmiller:
@bartblast This is not yet fully ready for merge.  It could use a number of improvements, certainly as pertain to documentation but maybe also correctness.  I wanted to create the PR, though, so that you could provide any feedback on how those might proceed.  I'm trying to keep things in line with the code conventions I see around.  Let me know where I have failed on that front as well.

mwmiller:
This has some issues when I apply the `plugin:` to the code in the test directory.  Some notes, mostly for myself:

- `~HOLO""` causes some heartburn.  The sigil is ineterpeted as `~HOLO"` with a text `"` in the body.
- The insertion of a "nice" `\n` at the end of a template doesn't always apply.  Need to look into making that bit smarter.

mwmiller:
I made a change about which I am not overly excited to make the line endings stuff work better for triple quoted strings. It works for my project's templates, but I want to do more to verify its correctness.

As part of my testing, I also added the `plugin` to this repository.  It doesn't actually change an of the extant sigils, so that's somewhat gratifying.

mwmiller:
I resolved the outstanding issues.  I am still dickering about the exact formatting and (somewhat dependently) configuration options.

I am also not sure about the best way to apply testing without making a mess of your repo.

mwmiller:
> * `~HOLO""` causes some heartburn.  The sigil is ineterpeted as `~HOLO"` with a text `"` in the body.

This was the same as the problem which followed it.  I was injecting a newline into an empty template... of course.

bartblast:
Hey @mwmiller, awesome work, thanks! :)

Initial thoughts:
* I like the state machine approach - that's a solid foundation for this kind of parsing work
* Are there things we should add/update in the parser itself to make formatting easier? For example, adding metadata to parser output that could help the formatter?
* You mentioned "not sure about the best way to apply testing without making a mess of your repo" - I'm not sure I understand why tests would make a mess. Can you clarify what you're concerned about?
* Would love to see some basic tests to understand how it's working in practice

mwmiller:
> * I like the state machine approach - that's a solid foundation for this kind of parsing 
work

Thanks, it was something of your idea and working with the output of parser makes it come out pretty well.

> * Are there things we should add/update in the parser itself to make formatting easier? For example, adding metadata to parser output that could help the formatter?

I am not sure if *within the parse output itself* there is anything which would be *especially* useful.

My biggest hassle (as I have mentioned to you elsewhere) is that the `{ else }` tags count as `:block_start` without a corresponding `:block_end`.  This requires some special casing in the formatter to say "indented just like the containing `if` and its contents as well"

The major concern with the above is what if there are other "sub-block" constructs lurking.  There's a literal hardcoded "else" in the formatter right now.  If the Parser had a way to programmatically say "`else` is always a sub-block of `if` or `with`.  `rescue` is always a sub-block of `try`." or whatever, then I would feel better about the reduced fragility.

It's also worth noting that the Formatter breaking is far from a critical path concern.  It might be annoying if people come to like and use it, but it won't destroy careers.

> * You mentioned "not sure about the best way to apply testing without making a mess of your repo" - I'm not sure I understand why tests would make a mess. Can you clarify what you're concerned about?

The entire point of writing Formatter is to have a nice integration with `mix format`.  To truly demonstrate its worth, you want to (as closely as possible) run `mix format` on a known "malformed" template and get back a "well-formed" one.

Doing this in inside the repository means I either need to clutter up `test_helper.exs` with something to create/revert malformed files or do some kind of testing which isn't on a `~HOLO` template in an Elixir source file.  Somewhere in the back of my mind, I think it is also possible that I might mess up the "known good" with the Formatter making it test bad formatting against bad formatting.

> * Would love to see some basic tests to understand how it's working in practice

Sure thing.  I've been busy with the start of the (American) football season.  I'll get back to this within the next couple of days.  I'll work on adding some test cases which demonstrate its value and limitations.  That should pick a path forward.

mwmiller:
I started writing a test of the `format/2` function just to see how it might look.  It's not so bad.

As one might hope, the first test case I wrote managed to irritate me with the output.  So, I guess I'll start doing very delayed TDD against some edges I can identify.

It should be possible to follow the tests herein to see how development is progressing.  I imagine a week before I feel comfortable submitting it for review.

mwmiller:
@bartblast I fixed the thing annoying me above and created a few tests for one to get a feel for how it would work.  Before I continue on with more some questions for your input:

- If I see `<SelfClosingTag/>` it becomes `<SelfClosingTag />`.  Is this the kind of opinion which should be a configuration option (like the two spaces indent)?
- Everything I do is adjusting whitespace. In places where I don't **know** if the whitespace is significant (notably expressions), I just leave things untouched.  Are there any land mines of which I should be aware?  Or places where it seems likely I could improve the output visually by taking a more aggressive stance? (NB: I have yet to have a template stop compiling or appear (to me) visually different when rendered.
- I'm using this hand constructed list of "block" tags.  That just means that it expects it will contain a bunch of sub-tags and should be set off on its own line.  Such a bespoke list is going to have a lot of omissions.  Any suggestions on how to find/maintain an authoritative list?  (I'm going to check the Phoenix `~H` sources soon. ETA: I looked and they do this the "opposite" way with a list of `inline` tags.  I am going to consider this.)

More tests coming in the next couple of days but any help you can give with the above is welcome for directing my focus.

mwmiller:
I thought I had looked at `Inspect.Algebra` previously and found it was insufficiently flexible.  I now believe I must have been looking in the wrong place or at the wrong thing.  I'd prefer to leverage that by building up documents from the parse and allowing it to render the text.

mwmiller:
> I'd prefer to leverage that by building up documents from the parse and allowing it to render the text.

I've started on this a very tiny bit and I believe it will make my life easier.  It doesn't at present, but that's fine.  I might get to this over the weekend and have something suitable thereafter.

mwmiller:
`Inspect.Algebra` doesn't feel like it is going to get me where I want to be.  I feel like I have to over-specify for the fairly simple needs I have.

I ran the current version over my Hologram project and identified some more failings.  I will develop some tests related thereto and iterate further tomorrow.

bartblast:
Hey @mwmiller! Thanks for the continued work on this!

Some thoughts on your questions and the direction:

**Scope for v1:** I think for the first version it could be a good idea to just focus on indentation and attributes/props formatting. Once we've got that working on solid foundations, we can build on it incrementally.

**Expressions:** I wouldn't touch expressions at all for now - you're right to leave whitespace untouched there since we can't be sure about significance.

**Configuration:** To be honest, I'm not sure we need configuration options. That's one of the pros of `mix format` - it stopped pointless discussions about conventions and left it to the community. What do you think about keeping it opinionated?

**Self-closing tags:** I think we should go with `<MyComponent />` (space before slash) - that seems to be the standard convention.

**Architecture idea:** What do you think about introducing a `DOMTree` module that would be responsible for building a document tree? This would make it easier to reason about the code, and if there are future changes to the template syntax, the formatting rules would be much easier to update.

Besides, if we have a DOMTree building step, it could also be used for future tooling like linters, template analyzers, or IDE extensions. The `Hologram.Template.DOM` module could potentially be simplified this way too.

So the pipeline would be: `Tokenizer -> Parser -> DOMTreeBuilder -> Formatter`

The tree approach would help with the if/else blocks/subblocks problem you mentioned - you'd be able to treat the whole construct as a node in the tree and work on it recursively. This approach could be more future-proof if there are template engine changes.

I had planned to implement something like a DOM tree builder myself, but honestly I followed the path of least resistance there 😅

**Block/inline elements:** I don't know of any authoritative list of block/inline tags, but if there is one, it's probably somewhere in the [HTML spec](https://html.spec.whatwg.org/). 

You can also take a look at the [Phoenix LiveView formatter](https://github.com/phoenixframework/phoenix_live_view/blob/c082d43e800f0b042e01f2580e32549b950ef1fb/lib/phoenix_live_view/html_formatter.ex#L184) - they keep a list of inline elements from what I understand.

I haven't dug deep into your code yet, so this is just brainstorming at this point. What are your thoughts on the DOMTree approach?

mwmiller:
> **Scope for v1:** I think for the first version it could be a good idea to just focus on indentation and attributes/props formatting. Once we've got that working on solid foundations, we can build on it incrementally.

Sure thing.  It is working well-enough for my personal project today.  The resulting HTML subjectively "looks better" and "doesn't break."  Those are my two criteria for this project, so that's sweet!

> **Expressions:** I wouldn't touch expressions at all for now - you're right to leave whitespace untouched there since we can't be sure about significance.

Sure thing.  I do mangle some whitespace in `:text` nodes, but I leave `:expressions` alone.  I **do** have to touch them in one place.  The conditional expression in the `Parser` output of an `if` block is enclosed in `{}`.  These don't exist in the actual source, so I strip the out in what I consider a [dumb way](https://github.com/bartblast/hologram/pull/250/files#diff-028b0c7ec80f60645ce33437dabbf3644bceba0acb2ed48df6970d7e00c62eafR171-R177).

> **Configuration:** To be honest, I'm not sure we need configuration options. That's one of the pros of `mix format` - it stopped pointless discussions about conventions and left it to the community. What do you think about keeping it opinionated?

I'm in favor.  Even when a formatter author makes the "wrong" choice it saves me a bunch of puttering around in making my own choice.  So this is the answer for which I was hoping.  The formatter does get the options from `.formatter.exs` if there are any we ought to respect.  This project has an extensively modified `.formatter.exs`. I poked through and didn't see anything about which I should concern myself (especially since I leave expressions untouched.)

> **Self-closing tags:** I think we should go with `<MyComponent />` (space before slash) - that seems to be the standard convention.

Cool.  This is already in there, then.  I might also remove the `holo_indent` configuration  option (default: `"  "`).  Two space indentation seems pretty standard in the community.

> **Block/inline elements:** I don't know of any authoritative list of block/inline tags, but if there is one, it's probably somewhere in the [HTML spec](https://html.spec.whatwg.org/).

I ended up using a list from MDN.  It has to do with what **renders** as block/inline in a browser.  It turns out this is a decent enough proxy for how you want your source structured as well.  I added and removed a couple tags based on visual review.

> You can also take a look at the [Phoenix LiveView formatter]

I have looked that code over a bit (particularly while on my `Inspect.Algebra` goose chasing. Their approach to blocking is approximately the opposite of mine. They keep a list of `inline` elements which shouldn't break; I keep a list of `block` elements which should. I (perhaps obviously) think my approach is better, though they are both kinda "meh" and fragile.  (On a callback note, they allow configuring the formatter with a different set of  `inline` elements which inspired some earlier questions.)

> I haven't dug deep into your code yet, so this is just brainstorming at this point. What are your thoughts on the DOMTree approach?

This was my original approach, actually.  It had some problems; perhaps some were self-inflicted.  

Probably the most notable one was that "templates don't have to be HTML documents."  They don't even have to be HTML tree fragments.  It's possible for any given template to be "unrooted", coming out of the parser with a (non-whitespace) `:text` node.  I spent some time wrapping in a fake root and treating it like a tree.  It felt pretty hacky

As you mentioned, there is also the issue of, say `{%if ...}` conditionals.  Is that contained block in the tree?  What if there is an `{else}`?  Are those two siblings in the same tree, even though they will never be rendered together? You mention treating it as a single node but I can imagine complex structures wherein this gets even messier.

For source formatting purposes, I can (mostly) ignore the HTML rendering and just worry about the source rendering.  Once I try to put it into a "working" DOMTree, I have problems with "validating" that tree.

mwmiller:
I've been adding some test cases which further irritate me about both the output and code structure.  I have a plan to improve both with some applied effort tomorrow.

I am wondering if "control blocks" (like `{%if ...}`) should also be "source blocks" where the contained structure is set off with line breaks and indentation.  My first instinct is "of course, it's right in the name!", but I also can see where it might be something like:

```
<h3>Hello, {%if @user}{@user}{%else}guest{/if}</h3>
```

I would personally write this actual example something more like {@user || "guest"}, but it could be more complex with potential `nil`s or something.  I doubt if anyone would want that to suddenly become:

```
<h3>Hello, 
  {%if @user}
    {@user}
  {%else}
    guest
  {/if}</h3>
```

The first thing I am going to do tomorrow is look (again and more deeply) into `Inspect.Algebra`.  I feel like I must not be "getting it." It should be able to do more than I have found.

Regardless, I do want to get "something" to you sooner rather than later to get this off my plate and into someone else's hands.

bartblast:
> > **Expressions:** I wouldn't touch expressions at all for now - you're right to leave whitespace untouched there since we can't be sure about significance.
> 
> Sure thing. I do mangle some whitespace in `:text` nodes, but I leave `:expressions` alone. I **do** have to touch them in one place. The conditional expression in the `Parser` output of an `if` block is enclosed in `{}`. These don't exist in the actual source, so I strip the out in what I consider a [dumb way](https://github.com/bartblast/hologram/pull/250/files#diff-028b0c7ec80f60645ce33437dabbf3644bceba0acb2ed48df6970d7e00c62eafR171-R177).

Do you mean leaving this untouched: `<MyWidget extra={ @nonsense } plus={@more} />`?
Actually, we should trim the whitespace inside the outer curly braces, because they're treated as regular tuples by the Hologram compiler with special behavior depending on what they contain. They're tuples in essence, so they should be formatted as tuples.

---

> I might also remove the `holo_indent` configuration option (default: `" "`). Two space indentation seems pretty standard in the community.

👍 

---

> Probably the most notable one was that "templates don't have to be HTML documents." They don't even have to be HTML tree fragments. It's possible for any given template to be "unrooted", coming out of the parser with a (non-whitespace) `:text` node. I spent some time wrapping in a fake root and treating it like a tree. It felt pretty hacky

Yeah, templates can be just a list of nodes - this is actually more flexible than LiveView's requirement for a single root node, which often feels artificial. I'm curious about your concern with the "virtual" root node approach though. What specifically makes it feel hacky to you? Is it the temporary nature of it, or something about how it affects the formatting logic?

---

> As you mentioned, there is also the issue of, say `{%if ...}` conditionals. Is that contained block in the tree? What if there is an `{else}`? Are those two siblings in the same tree, even though they will never be rendered together? You mention treating it as a single node but I can imagine complex structures wherein this gets even messier.

When I mentioned "DOM tree", I meant our document object model tree, not HTML specifically. Thinking about it this way, any structure that evaluates to an expression can be treated as a node in the tree. Looking at how Elixir code is generated from templates, it's just a tree structure anyway. So for `if` blocks, we could have something like `%IfBlock{do: ..., else: ...}` (or any other pair like true/false, then/else, etc.).

---

> I am wondering if "control blocks" (like `{%if ...}`) should also be "source blocks" where the contained structure is set off with line breaks and indentation. My first instinct is "of course, it's right in the name!", but I also can see where it might be something like:
> 
> ```
> <h3>Hello, {%if @user}{@user}{%else}guest{/if}</h3>
> ```
> 
> I would personally write this actual example something more like {@user || "guest"}, but it could be more complex with potential `nil`s or something. I doubt if anyone would want that to suddenly become:
> 
> ```
> <h3>Hello, 
>   {%if @user}
>     {@user}
>   {%else}
>     guest
>   {/if}</h3>
> ```

I don't think we should format control blocks like `{%if ...}`. You correctly noticed that it could introduce unintended whitespace inside the branches, but the developer wrote it as a single line intentionally. We should respect their formatting choice and only format in cases when it improves readability without changing the developer's intent.

mwmiller:
@bartblast Sorry for the delay in responding.  I've had soe unfortunately family issues and a bit of work-like stuff to complete.

I'm onboarding your advice as I progress, likely tomorrow.

> What specifically makes it feel hacky to you? Is it the temporary nature of it, or something about how it affects the formatting logic?

Well, how I was doing it was to introduce a faux `holo_root` tag as the tree base, then building a up underneath it... only to add the "closing" `holo_root` at the end... and ignore those tags in the actual formatting.  So a whole lot of indirection for little to no value.

There are still remnant of this approach in my code which I hope to eliminate on this next pass.  I had a sort of epiphany on how to improve my indentation conditionals last time I looked at this.  I expect to be able to reduce the code base line count and complexity significantly.  We'll see if that comes to fruition. :grin:

mwmiller:
I just pushed a new single commit which greatly simplifies things.  I don't have time to look into some of the new edges introduced.  I also haven't taken care of the expression squeezing just yet.  I expect to have time to work on that tomorrow or the next day.

ETA: I fixed up the code breaking tests in the conversion and added the expression squishing.  It now produces the agreed and desired output for everything tested. I expect to spend tomorrow morning writing more tests to search for unsightly edges.  Maybe I'll actually be able to deliver this!  Simply running it against templates in the repo itself demonstrates some failings, so it should "easy."

mwmiller:
I made a lot of progress this morning without writing any tests.  My focus is on maintaining developer intent.  This makes the results somewhat non-deterministic.  If I think you intended to introduce some newlines as part of an aesthetic choice, then I put in a "properly indented" newline.

This cause much less pain when I run `mix format` against this repository.  When I have some more time, I am going to crib some test cases from these "edges."  As a note mostly to myself, one big thing is working through conditional block handling.  I believe I am going to need more state to handle this.

mwmiller:
I added some tests and made some more progress.  I also realized I may have broken some semantics with my whitespace trimming.  I don't have time to fix it right now, so I will make that my first task on my next go 'round.

mwmiller:
This now does mostly what I want.  I need to look further into divining developer intent around conditional blocks.

It also fails rather spectacularly on `lib/hologram/ui/runtime.ex`.  This file has a templated `<script>` tag which gets mangled and becomes uncompilable.  I tried adding an `inviolable` tracker for `<script>` (and potentially other future) tags.  It works but produces poor output.  So I'm going to let that percolate.

Over the rest of the repo it changes 17 other files.  Some in ways with which I agree and others in ways which I find "acceptable"... but might mask developer intent.

So, I'll be back at this in a few days when I have time and a hopeful epiphany.

mwmiller:
I had some more time with this today.  I tried (again) to create an `inviolable` section when something is wrapped in `<script>` tags.  However, looking more closely at the `Parser` output for `lib/hologram/ui/runtime.ex` it doesn't contain all of the source information.

I can certainly skip this file in the formatter, but I want to make sure it is an intentional `HOLO` special-case.  I still have lots to work out for making it friendly, so this isn't blocking.

mwmiller:
It seems like it might be only the `{%raw}` blocks missing from the parser.  A minimal example from the repo (`test/features/app/pages/template_syntax/raw_block_page.ex`):

```
 ~HOLO"{%raw}{%if false}abc{@var}xyz{/if}{/raw}"
```

Parses to:
```
[text: "{%if false}abc{@var}xyz{/if}"]
```

I haven't looked too deeply, but it looks like this is to allow for a macro-esque building of a  `HOLO` template inside a `HOLO` template.  This doesn't confront me much either way, but the loss of information from the parse makes it essentially impossible to reproduce the source directly.

I'm going to poke further at the Parser to see if there are other subtleties I have missed.

ETA: It seems like the only other major mode of the Parser is `script`.  It appears to do exactly what I am trying to do on the other side of the parse: maintain everything inside the `<script>` tags as inviolable text.  So that's not an issue in the formatter context.

bartblast:
Hey Matt,

Is the missing `{%raw}` block information in the parser output the main blocker at this point, or are there other issues you're encountering as well (like with `lib/hologram/ui/runtime.ex`)?

If the raw blocks are indeed the main issue, what do you think should be the course of action here? Should I update the parser to preserve that information, or would you prefer to handle it differently in the formatter?

mwmiller:
> Is the missing `{%raw}` block information in the parser output the main blocker at this point, or are there other issues you're encountering as well (like with `lib/hologram/ui/runtime.ex`)?

`raw` is the issue which makes this wholly untenable at this point.  I see two ways to proceed.  Either have the parser include the `raw` block tags or do a static inspection of the source before passing it to the parser.  In this second case, I would simply skip formatting of template sources.  I could add documentation or a warning.

(NB. The `Mix.Tasks.Format` behaviour does not include source attributes to indicate "skip this file/section."  I'm vaguely considering making an issue/PR there because current best practice seems... sub-optimal.)

The other *major* issue is that I need to make `<script>` (and other?) sections inviolable so that I don't accidentally adjust code which appears to the formatter as `:text`.  Presently this actually causes the repo source to become uncompilable after format.  Working on this is what surfaced the `raw` issue for me initially.

There are also some minor issues with "respecting developer intent."  The first example which shows in the current output:

```
-    <a 
-      href={page_path(@to)}
-      class={@class}
-      rel={@rel}
-      style={@style}
-      $pointerdown={:__prefetch_page__, to: @to}
-      $click={:__load_prefetched_page__, to: @to}
-    ><slot /></a>
+    <a href={page_path(@to)} class={@class} rel={@rel} style={@style} $pointerdown={:__prefetch_page__, to: @to} $click={:__load_prefetched_page__, to: @to}><slot /></a>
```

You obviously intend to have the attributes on separate lines, but that's not "normal" behavior for the formatter.  These kind of line-length rethinkings are why I was going to switch to `Inspect.Algebra`.  That turned out to be a dead end for reasons I cannot recall at the moment.

I've also still got some spacing and more developer intent stuff to work out.  A current example:

```
-    layout template start
-    <Runtime />
-    <slot />
-    layout template end
+    layout template start<Runtime /><slot />layout template end
```

This unrooted fragment loses the spacing between the text sections and self-closing tags.  Also, whether the self-closers should be on their own lines.

mwmiller:
I spent some more time looking at the Phoenix HTMLFormatter. I see that they are using two passes. This allows for proper contextual lookahead/back and makes the use of `Inspect.Algebra` much more sane. I think I have been spending too much time trying to make everything work in a single pass for some kind of false efficiency

So, I'm going to go ahead and switch to two passes and see where that leads me.

For the record that two passes will be:

- Convert parse into tree(s).
- Apply `Inspect.Algebra` to those trees

I promise, I will (eventually!) produce a working parser.  😁

bartblast:
> `raw` is the issue which makes this wholly untenable at this point. I see two ways to proceed. Either have the parser include the `raw` block tags or do a static inspection of the source before passing it to the parser. In this second case, I would simply skip formatting of template sources. I could add documentation or a warning.

This shouldn't be a problem - feel free to add the raw block tags similar to how other blocks are flagged in the parser output. Currently `:raw?` is just an opaque internal flag in the parser context, but you can make it explicit by outputting block tags.
Looking at the parser, you'll see that `{%if}`, `{%for}`, and `{%else}` all produce explicit tags in the output:
* `{:block_start, "else"}`
* `{:block_start, {"for", expression}}`
* `{:block_end, "for"}`

Right now `{%raw}` and `{/raw}` (lines 646-658 in parser.ex) only call `enable_raw_mode()` and `disable_raw_mode()` which toggle the internal `:raw?` flag, but they don't emit any block tags. You could probably follow the same pattern as if/for blocks - something like this (just an idea from the top of my head, haven't verified it, there could be more to this):

```elixir
def parse_tokens(%{raw?: false} = context, :text, [{:symbol, "{%raw}"} = token | rest]) do
  context
  |> maybe_add_text_tag()
  |> add_processed_token(token)
  |> add_processed_tag({:block_start, "raw"})  # Add this
  |> enable_raw_mode()
  |> parse_tokens(:text, rest)
end

def parse_tokens(%{raw?: true} = context, :text, [{:symbol, "{/raw}"} = token | rest]) do
  context
  |> maybe_add_text_tag()
  |> add_processed_token(token)
  |> add_processed_tag({:block_end, "raw"})  # Add this
  |> disable_raw_mode()
  |> parse_tokens(:text, rest)
end
```

bartblast:
> * Convert parse into tree(s).

Is this the DOM tree I mentioned earlier or something else?

bartblast:
> The other _major_ issue is that I need to make `<script>` (and other?) sections inviolable so that I don't accidentally adjust code which appears to the formatter as `:text`. Presently this actually causes the repo source to become uncompilable after format. Working on this is what surfaced the `raw` issue for me initially.

if you have the DOM tree structure, you can traverse the tree and know that you're currently inside a script tag and act accordingly. The tree would have the hierarchical structure, so when you're processing text children of a script element, you'd know to leave them untouched. This should solve the issue of text inside `<script>` tags being accidentally reformatted.

mwmiller:
> You could probably follow the same pattern as if/for blocks 

Ok so long as you (and the other consumers) don't mind, I'll probably do that.  I can also clean up any detritus which might come from the extra emissions.

> Is this the DOM tree I mentioned earlier or something else?

This is essentially the DOM tree about which we spoke.  In a surprising number of cases is ends up being a mini-forest, as the document fragments are unrooted.  This should end up being fine.  The nesting happens a bit naturally which should allow for easy Algebra.  (NB I started on this but haven't made significant progress.  I had some compilation and dependency issues surprise and nerd-snipe me.)

> This should solve the issue of text inside <script> tags being accidentally reformatted.

Yep, this is another advantage of the two-pass system.  I don't need to track as much state on the output side.  Simply "leave untouched if inside <script>/<style>/<other non-html-containing tags>"

mwmiller:
> (just an idea from the top of my head, haven't verified it, there could be more to this):

I had a little bit of time to work on this today.  There is, indeed, more to it, but seemingly not a lot. My biggest issue is getting double text nodes when inserting the `raw` blocks.  I assume this is related to `maybe_add_text_tag/1`.  I haven't yet read it, but it behaves in a way which surprises me at this point.

The big breakage is in the `DOM` AST generation.  I think `raw` is only useful for source inspection/manipulation so I added a `{:block_start/:block_end, "raw"}` render_code handler which expresses `:skip` and put a `case` into the `reduce` block which calls `render_code`.  I think this is fine if it doesn't offend your sensibilities. We could theoretically skip more source annotations, but I don't know whence they would come.

Finally, it will require a bunch of changes for tests which examine the `Parser` output directly.  A lot of that is macro'd in place so it should be fairly simple to update them.

I plan to submit these changes as a precusor PR before continuing with those changes as a base in this PR.

ETA: I stayed up past my bedtime to see the tests passing.  I have some (mostly style-matching) clean up to do when I'm more awake.  I expect to get the other PR in the queue tomorrow. (Any further discussion on this can appear in #287 so as not to further clutter this PR with my ramblings.)

mwmiller:
Closing this in favor of another PR to come later.  It builds some on this, plus the raw changes and `Inspect.Algebra`.  It is partially working now.

The changes are so extensive that it makes sense to start afresh without all of my yapping herein as a distraction.

PR #281, opened 2025-10-20
Improve compile speed with Agent adjustment

mwmiller:
Move Agent.update/3 calls to Agent.cast/2

Since the compiler acts on an isolated node and deals with single serializing processes, it is safe to run our mutations asynchronously. Subsequent `get/3` calls will appear later in the process mailbox and see the final mutated state properly.

This has some measurable performance impact on my local setup.

All tests pass locally.

@bartblast I didn't update any benchmark outputs.  My setup differs significantly the one used in your established baselines.

### DIFF
--- a/lib/hologram/compiler/call_graph.ex
+++ b/lib/hologram/compiler/call_graph.ex
@@ -140,7 +140,7 @@ defmodule Hologram.Compiler.CallGraph do
   """
   @spec add_edge(t, vertex, vertex) :: t
   def add_edge(%{pid: pid} = call_graph, from_vertex, to_vertex) do
-    Agent.update(pid, &Digraph.add_edge(&1, from_vertex, to_vertex), :infinity)
+    Agent.cast(pid, &Digraph.add_edge(&1, from_vertex, to_vertex))
     call_graph
   end
 
@@ -150,7 +150,7 @@ defmodule Hologram.Compiler.CallGraph do
   """
   @spec add_edges(t, [edge]) :: t
   def add_edges(%{pid: pid} = call_graph, edges) do
-    Agent.update(pid, &Digraph.add_edges(&1, edges), :infinity)
+    Agent.cast(pid, &Digraph.add_edges(&1, edges))
     call_graph
   end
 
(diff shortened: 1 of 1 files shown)


bartblast:
Thank you @mwmiller! :)

The optimization looks sound  - message processing guarantees make `Agent.cast` safe here, and the linked Agent ensures errors still crash the compiler.

One trade-off to note: Error messages will be less clear. With `Agent.update`, exceptions in Digraph operations raise directly with full stack traces. With `Agent.cast`, you'll just see "Agent crashed" without specific operation context. Given that Digraph functions are simple Map operations, this is probably acceptable.

I've improved the benchmark script to generate summaries that make baseline comparisons easy: https://github.com/bartblast/hologram/tree/dev/benchmarks/elixir

I've merged the PR, will re-run the benchmarks locally, and will test the feature app with the new version to make sure everything works correctly (can't update the feature app's Hologram dep from an external branch). I'll report the results back here.

mwmiller:
> you'll just see "Agent crashed" without specific operation context. 

I did notice this in my ill-considered first pass at this.  My thinking was that this is so deeply internal that the end user is unlikely to have any recourse outside of "re-run", anyway.  Also, I couldn't *actually* figure out how to induce a failure.  As you say, it's pretty simple internally.

bartblast:
Looks good! The feature tests passed, and the performance improved.

https://github.com/bartblast/hologram/commit/b4c898bd8ef459141201ceff336da679dda450ad#diff-f99882a18f4b376bc1c4cdf7cdfb6dbbd242b205ea9d9bd059326a7b4b67b86e

### ✅ compiler » build_call_graph_1

ips | average | deviation | median | 99th %

OLD:
1.62
618.39 ms
±2.04%
614.92 ms
651.85 ms

NEW:
3.71
269.87 ms
±3.45%
268.73 ms
311.06 ms

mwmiller:
In a real world test, my application compilation speed went from ~6s to ~4s.  It doesn't look like much, but it's palpable in the reload cycle.

PR #287, opened 2025-10-31
Emit :raw blocks from Parser

mwmiller:
These tags have no semantic value in the AST, so they are elided from that output.

Including them in the Parser output makes it possible to reconstruct the source more correctly from said parse.  This is valuable when dealing with the templates as source text, such as in the nascent Formatter module.

In the AST builder, `render_code` is updated to `:skip` "raw" tags. The most likely source of error here is handling a `[text: "raw"]` node. Tests are added to demonstrate probable correctness.

This `:skip` arrangement makes it easy to add any future "source-only" code handling, even though it is hard to see whence that would come.

A `wrap_in_raw` test helper is created to reduce the diff size of this change.  It demonstrates more clearly how the `raw` tag wraps (portions of) the output.  This can easily be revisited if it is felt that the more human-friendly option is having the literal output inline.

### DIFF
--- a/lib/hologram/template/dom.ex
+++ b/lib/hologram/template/dom.ex
@@ -30,10 +30,14 @@ defmodule Hologram.Template.DOM do
     {code, _last_tag_type} =
       Enum.reduce(tags, {"", nil}, fn tag, {code_acc, last_tag_type} ->
         current_tag_type = if is_tuple(tag), do: elem(tag, 0), else: tag
-        current_tag_code = render_code(tag)
-        new_code_acc = append_code(code_acc, current_tag_code, last_tag_type)
+        # :skip items are fully elided, as if they did not appear
+        case render_code(tag) do
+          :skip ->
+            {code_acc, last_tag_type}
 
-        {new_code_acc, current_tag_type}
+          current_tag_code ->
+            {append_code(code_acc, current_tag_code, last_tag_type), current_tag_type}
+        end
       end)
 
     "[#{code}]"
@@ -62,6 +66,17 @@ defmodule Hologram.Template.DOM do
     expr_str
     |> String.slice(1, String.length(expr_str) - 2)
     |> String.trim()
+  end
+
+  # `raw` blocks are useful only in the handling of template sources.
+  # `Parser` emits them so that such source can be reconstructed.
+  # They can be skipped in the building of the AST here.
+  defp render_code({:block_start, "raw"}) do
+    :skip
+  end
+
+  defp render_code({:block_end, "raw"}) do
+    :skip
   end
 
   defp render_code({:block_start, "else"}) do
--- a/lib/hologram/template/parser.ex
+++ b/lib/hologram/template/parser.ex
@@ -94,7 +94,7 @@ defmodule Hologram.Template.Parser do
             delimiter_stack: list(delimiter),
             node_type: :attribute | :block | :public_comment | :tag | :text,
             prev_status: Parser.status() | nil,
-            processed_tags: [],
+            processed_tags: list(Parser.parsed_tag()),
             processed_tokens: list(Tokenizer.token()),
             raw?: boolean,
             script?: boolean,
@@ -646,13 +646,17 @@ defmodule Hologram.Template.Parser do
   def parse_tokens(%{raw?: false} = context, :text, [{:symbol, "{%raw}"} = token | rest]) do
     context
     |> add_processed_token(token)
+    |> add_processed_tag({:block_start, "raw"})
     |> enable_raw_mode()
     |> parse_tokens(:text, rest)
   end
 
   def parse_tokens(%{raw?: true} = context, :text, [{:symbol, "{/raw}"} = token | rest]) do
     context
+    |> maybe_add_text_tag()
+    |> reset_token_buffer()
     |> add_processed_token(token)
+    |> add_processed_tag({:block_end, "raw"})
     |> disable_raw_mode()
     |> parse_tokens(:text, rest)
   end
--- a/test/elixir/hologram/template/dom_test.exs
+++ b/test/elixir/hologram/template/dom_test.exs
@@ -1175,6 +1175,41 @@ defmodule Hologram.Template.DOMTest do
     end
   end
 
+  describe "build_ast/1, raw block" do
+    test "literal tag elided" do
+      parse = [
+        {:block_start, "raw"},
+        {:block_end, "raw"}
+      ]
+
+      assert build_ast(parse) == []
+    end
+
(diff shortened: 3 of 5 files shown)


mwmiller:
I tried to explain my changes well in the commit message for posterity.  I know I can be annoyingly verbose; so, sorry about that.

I expect that the main point of contention might be adding the test helper instead of updating the tests with literal lines.  I am happy enough to go add the lines by hand if we feel that it would be more valuable to future readers.

bartblast:
Hey Matt,
Thanks for the PR! The approach makes sense for the formatter work.

A few things to adjust:
* **wrap_in_raw/1 helper**: Could you use explicit tags instead? Since we're testing actual parse output in isolation, it would be clearer to see what's actually there, and it'd be more consistent with the rest of our tests.
* **dom_test.exs**: Since you modified the `dom.ex` module, it makes sense to cover those changes in `dom_test.exs` as well.
* **Integration test**: For the `"template with raw section"` test in `template_test.exs`, it would be better to use some simple template syntax that should be disabled instead of the text `"raw"` - something like `{%if true}Hologram{/if}` or whatever makes sense. This would demonstrate the actual use case better.

Also, there's a [Credo warning](https://github.com/bartblast/hologram/actions/runs/18964696243/job/54158922612?pr=287) showing up, but it's related to the `wrap_in_raw helper/1` which will be removed anyway, so no worries there.

Thanks again!

mwmiller:
I've pushed changes in line with what you've asked above.

Apologies about the credo thing.  I haven't used it in quite some time because it would produce spurious warnings which made things worse or harder to read.  (This kinda fits that since I actually know the size of the list never exceeds 4, but it cannot know that obviously.)

[Edit: After writing the above I figured out what I thought were weird LSP syntax warnings from `elixir-tools` are actually Credo diagnostics.  I installed `trouble.nvim` to improve their visibility.  Credo has gotten much better than I remember; thanks for having it in the project!]

I also apologize for not finding the `dom_test` initially (or recognizing that `template_test` was integration.)  I don't know what happened to my `rg` abilities this morning!

bartblast:
No worries at all!
CI tests/checks run automatically on every PR, so no need to think about it. Just make sure they pass and fix any that fail.

Credo can be too strict sometimes. For cases where it's too much, I usually use credo:disable comments. Generally they're reasonable, though I did remove some default checks in the past.

Regarding `add_processed_tag/2`: I'd like to keep the previous version because it was more readable and slightly faster. Since you changed it to fix a Dialyzer warning, could you share the Dialyzer output? I can't test it locally with this external PR.

mwmiller:
Dialyzer errors:

```
lib/hologram/template/parser.ex:650:8:call
The function call will not succeed.

Hologram.Template.Parser.parse_tokens(
  %{:processed_tags => nonempty_maybe_improper_list(), :raw? => true, _ => _},
  :text,
  _rest :: any()
)

breaks the contract
(Hologram.Template.Parser.Context.t(), status(), [Hologram.Template.Tokenizer.token()]) ::
  [parsed_tag()]

________________________________________________________________________________
lib/hologram/template/parser.ex:660:8:call
The function call will not succeed.

Hologram.Template.Parser.parse_tokens(
  %{:processed_tags => nonempty_maybe_improper_list(), :raw? => false, _ => _},
  :text,
  _rest :: any()
)

breaks the contract
(Hologram.Template.Parser.Context.t(), status(), [Hologram.Template.Tokenizer.token()]) ::
  [parsed_tag()]

________________________________________________________________________________
```

I originally misinterpreted this output as complaining about the `_rest` not being `Hologram.Template.Tokenizer.token()` which was even weirder.  On further inspection, it became clear that the structure was being typed as a plain map.  I updated the `@spec` to accept `term` in the first position and everything passed.

Obviously, I didn't want to change the actual contract, so this was the most evident way to demonstrate problem and potential solution.  (Note also that wrapping the `%{context|}` update in `IO.inspect` made dialyzer pass.) Additionally, the `add_processed_token/2` which uses the exact same construct never fails.  Especially confusing as the `add_processed_tag/2` call checking only seems to fail in the code I added.

I clearly lack the right mental model of how dialyzer is working here.

mwmiller:
> I can't test it locally with this external PR.

The GitHub forking model really sucks for its supposed social coding purpose.  For the record, I expect at least a couple more PRs from my fork. It might be worth adding a remote for my fork to allow tracking my nonsense.  If you just want to poke at it without a whole lot of extra overhead:

```
git remote add mwm git@github.com:mwmiller/hologram.git
git fetch --prune mwm
git checkout mwm/emit_raw
```

bartblast:
Hey @mwmiller, here's how to fix the Dialyzer warnings (I also reverted changes in `add_processed_tag/2` and `add_processed_token/2`):

```
diff --git a/lib/hologram/template/parser.ex b/lib/hologram/template/parser.ex
index 5145da783..b39996c73 100644
--- a/lib/hologram/template/parser.ex
+++ b/lib/hologram/template/parser.ex
@@ -94,7 +94,7 @@ defmodule Hologram.Template.Parser do
             delimiter_stack: list(delimiter),
             node_type: :attribute | :block | :public_comment | :tag | :text,
             prev_status: Parser.status() | nil,
-            processed_tags: [],
+            processed_tags: list(Parser.parsed_tag()),
             processed_tokens: list(Tokenizer.token()),
             raw?: boolean,
             script?: boolean,
@@ -856,12 +856,12 @@ defmodule Hologram.Template.Parser do
     add_processed_tag(context, new_tag)
   end
 
-  defp add_processed_tag(context, tag) do
-    Map.update(context, :processed_tags, [tag], fn prev_tags -> [tag | prev_tags] end)
+  defp add_processed_tag(%{processed_tags: processed_tags} = context, tag) do
+    %{context | processed_tags: [tag | processed_tags]}
   end
 
-  defp add_processed_token(context, token) do
-    Map.update(context, :processed_tokens, [token], fn prev_tokes -> [token | prev_tokes] end)
+  defp add_processed_token(%{processed_tokens: processed_tokens} = context, token) do
+    %{context | processed_tokens: [token | processed_tokens]}
   end
 
   defp add_public_comment_end_tag(context) do
```

Notice the `processed_tags` type spec.

mwmiller:
> Notice the `processed_tags` type spec.

Hey, cool!  This change actually explains the issues I was seeing!  I'll apply this momentarily.

mwmiller:
@bartblast Applied.  Things seem to be working.  I combined all of the back and forth into a single commit for the purposes of mainline history.  None of the extra commits captured much in the way of insight.

bartblast:
Thank you Matt! :)

PR #303, opened 2025-11-16
Port :binary.at/2 to JS

### LINKED ISSUE #301: Port :binary.at/2 to JS
I'm starting working on this now.

ftes:
closes #301

### DIFF
--- a/assets/js/erlang/binary.mjs
+++ b/assets/js/erlang/binary.mjs
@@ -12,23 +12,28 @@ const Erlang_Binary = {
   // Start at/2
   "at/2": function (subject, pos) {
     if (!Type.isBinary(subject)) {
-      Interpreter.raiseFunctionClauseError(
-        Interpreter.buildFunctionClauseErrorMsg(":binary.at/2", arguments),
+      const msg = Type.isBitstring(subject)
+        ? "is a bitstring (expected a binary)"
+        : "not a binary";
+      Interpreter.raiseArgumentError(Interpreter.buildArgumentErrorMsg(1, msg));
+    }
+
+    if (!Type.isInteger(pos)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(2, "not an integer"),
       );
     }
 
-    if (!Type.isInteger(pos) || pos.value < 0n) {
-      Interpreter.raiseFunctionClauseError(
-        Interpreter.buildFunctionClauseErrorMsg(":binary.at/2", arguments),
+    if (pos.value < 0n) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(2, "out of range"),
       );
     }
 
     Bitstring.maybeSetBytesFromText(subject);
 
     if (pos.value >= subject.bytes.length) {
-      Interpreter.raiseArgumentError(
-        Interpreter.buildArgumentErrorMsg(2, "out of range"),
-      );
+      Interpreter.raiseArgumentError("argument error");
     }
 
     return Type.integer(subject.bytes[pos.value]);
--- a/test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
@@ -29,13 +29,13 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
       end
     end
 
-    test "raises FunctionClauseError when subject is nil" do
+    test "raises ArgumentError when subject is nil" do
       assert_error ArgumentError,
                    build_argument_error_msg(1, "not a binary"),
                    fn -> :binary.at(nil, 0) end
     end
 
-    test "raises FunctionClauseError when bitstring is not a binary" do
+    test "raises ArgumentError when bitstring is not a binary" do
       subject = <<1::1, 0::1, 1::1>>
 
       assert_error ArgumentError,
@@ -43,13 +43,13 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
                    fn -> :binary.at(subject, 0) end
     end
 
-    test "raises FunctionClauseError when position is nil" do
+    test "raises ArgumentError when position is nil" do
       assert_error ArgumentError,
                    build_argument_error_msg(2, "not an integer"),
                    fn -> :binary.at(@binary, nil) end
     end
 
-    test "raises FunctionClauseError when position is negative" do
+    test "raises ArgumentError when position is negative" do
       assert_error ArgumentError,
                    build_argument_error_msg(2, "out of range"),
                    fn -> :binary.at(@binary, -1) end
--- a/test/javascript/erlang/binary_test.mjs
+++ b/test/javascript/erlang/binary_test.mjs
@@ -39,70 +39,50 @@ describe("Erlang_Binary", () => {
       assertBoxedError(
         () => at(binary, pos),
         "ArgumentError",
-        Interpreter.buildArgumentErrorMsg(2, "out of range"),
+        "argument error",
       );
     });
 
-    it("raises FunctionClauseError when subject is nil", () => {
+    it("raises ArgumentError when subject is nil", () => {
       const subject = Type.nil();
-      const pos = Type.integer(0);
-
-      const expectedMessage = Interpreter.buildFunctionClauseErrorMsg(
-        ":binary.at/2",
-        [subject, pos],
-      );
 
       assertBoxedError(
-        () => at(subject, pos),
-        "FunctionClauseError",
-        expectedMessage,
+        () => at(subject, Type.integer(0)),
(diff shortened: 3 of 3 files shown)


assets/js/erlang/binary.mjs
```diff
@@ -0,0 +1,40 @@
...
+      );
+    }
+
+    if (!Type.isInteger(pos) || pos.value < 0) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":binary.at/2", arguments),
+      );
+    }
+
+    Bitstring.maybeSetBytesFromText(subject);
```
ftes, line 33: OK to modify input?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,297 @@
...
+    test "returns byte at position 0" do
+      assert :binary.at(<<5, 19, 72, 33>>, 0) == 5
+    end
+
+    test "returns byte at position 1" do
+      assert :binary.at(<<5, 19, 72, 33>>, 1) == 19
+    end
+
+    test "raises ArgumentError when position is out of range" do
+      assert_error ArgumentError, "argument error", fn ->
```
ftes, line 27: I don't understand why the error only contains `"argument error"` as text.
Whereas other `ArgumentError`s below have a message such as  `"is a bitstring (expected a binary)"` or `"out of range"`.

assets/js/erlang/binary.mjs
```diff
@@ -0,0 +1,40 @@
...
+const Erlang_Binary = {
+  // Start at/2
+  "at/2": function (subject, pos) {
+    if (!Type.isBinary(subject)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":binary.at/2", arguments),
+      );
+    }
+
+    if (!Type.isInteger(pos) || pos.value < 0) {
```
bartblast: This is going to be marginally faster:
pos.value < 0n

assets/js/erlang/binary.mjs
bartblast: Yeah, we're good, this doesn't modify the value. Both bytes and text fields are kept as a performance optimization, so that when the object is passed around and used in different contexts we don't have to convert between text and binary representations.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
bartblast: Sometimes Erlang error messages are inconsistent 🤷‍♂️

test/javascript/erlang/binary_test.mjs
```diff
@@ -0,0 +1,448 @@
...
+import Type from "../../../assets/js/type.mjs";
+
+defineGlobalErlangAndElixirModules();
+
+describe("Erlang_Binary", () => {
+  describe("at/2", () => {
+    const at = Erlang_Binary["at/2"];
+
+    it("returns byte at position 0", () => {
+      const binary = Type.bitstring([
```
bartblast: You can use Bitstring.fromBytes([5, 19, 72, 33]) in this and similar tests.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,297 @@
...
+  IMPORTANT!
+  Each Elixir consistency test has a related JavaScript test in test/javascript/erlang/binary_test.mjs
+  Always update both together.
+  """
+
+  use Hologram.Test.BasicCase, async: true
+
+  @moduletag :consistency
+
+  describe "at/2" do
```
bartblast, line 13: Could you simplify the tests? Many appear redundant - most are essentially testing the `<<>>` operator. I'd suggest focusing on a few key scenarios: a representative middle byte case, edge cases, and error conditions. This would provide adequate coverage without duplication.

bartblast (changes requested):

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
ftes: Interesting.

But then why even bother to construct a nice error in `binary.mjs`, if it's silently ignored anyway?


```js
        Interpreter.buildArgumentErrorMsg(2, "out of range"),
      );
```

bartblast: The goal is to have the same behaviour for both client and server, so we need to update this - I see that "raises ArgumentError when position is out of range" are inconsistent.

test/javascript/erlang/binary_test.mjs
```diff
@@ -0,0 +1,109 @@
...
+      const result = at(binary, Type.integer(1));
+      assert.deepStrictEqual(result, Type.integer(19));
+    });
+
+    it("returns last byte", () => {
+      const result = at(binary, Type.integer(3));
+      assert.deepStrictEqual(result, Type.integer(33));
+    });
+
+    it("raises ArgumentError when position is out of range", () => {
```
bartblast, line 36: This one is inconsistent with the server version.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,58 @@
...
+
+    test "raises FunctionClauseError when bitstring is not a binary" do
+      subject = <<1::1, 0::1, 1::1>>
+
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "is a bitstring (expected a binary)"),
+                   fn -> :binary.at(subject, 0) end
+    end
+
+    test "raises FunctionClauseError when position is nil" do
```
bartblast: The actual test shows ArgumentError

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,58 @@
...
+      assert_error ArgumentError,
+                   build_argument_error_msg(2, "not an integer"),
+                   fn -> :binary.at(@binary, nil) end
+    end
+
+    test "raises FunctionClauseError when position is negative" do
```
bartblast: The actual test shows ArgumentError

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+
+    test "raises FunctionClauseError when subject is nil" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "not a binary"),
+                   fn -> :binary.at(nil, 0) end
+    end
+
+    test "raises FunctionClauseError when bitstring is not a binary" do
```
bartblast: The actual test shows ArgumentError

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,58 @@
...
+      assert :binary.at(@binary, 3) == 33
+    end
+
+    test "raises ArgumentError when position is out of range" do
+      assert_error ArgumentError, "argument error", fn ->
+        :binary.at(@binary, 4)
+      end
+    end
+
+    test "raises FunctionClauseError when subject is nil" do
```
bartblast: The actual test shows ArgumentError

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
ftes: Can you point me to where to look/what to change to ensure consistency?

bartblast: First step: Ensure tests expect the same behavior.

This test:
```
test "raises ArgumentError when position is out of range" do
   assert_error ArgumentError, "argument error", fn ->
     :binary.at(@binary, 4)
   end
 end
```
is the baseline, since it reflects the server behavior.

However, in this test:
```
it("raises ArgumentError when position is out of range", () => {
    const pos = Type.integer(4);

    assertBoxedError(
      () => at(binary, pos),
      "ArgumentError",
      Interpreter.buildArgumentErrorMsg(2, "out of range"),
    );
 });
```
you're expecting "out of range" instead.

This means you need to raise ArgumentError with message "argument error" instead of "out of range" in the function.

Double-check that other client tests expect exactly the same behavior as the server tests - that's the goal of consistency tests. You want to ensure the client tests match the server tests exactly.

ftes: Thanks. I misunderstood the purpose of the consistency tests.
I thought they were integration tests, running the elixir code via hologram.

Should be correct now.

PR #318, opened 2025-11-16
Port :erlang.binary_part/3 to JS

### LINKED ISSUE #312: Port :erlang.binary_part/3 to JS

djonn:
Closes #312

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
@@ -1373,6 +1373,107 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "binary_part/3" do
+    test "first argument is a text binary" do
+      assert :erlang.binary_part("goldfish", 4, 4) == "fish"
+    end
+
+    test "first argument is a byte binary" do
+      assert :erlang.binary_part(<<1, 2, 3, 4, 5, 6, 7, 8, 9, 10>>, 4, 2) == <<5, 6>>
+    end
+
+    test "first argument is empty" do
+      assert :erlang.binary_part("", 0, 0) == ""
+    end
+
+    test "first argument contains multi-byte unicode characters" do
+      # The character "á" is represented in UTF-8 as two bytes: <<195, 161>>
+      assert :erlang.binary_part("á", 0, 1) == <<195>>
+    end
+
+    test "second argument is zero" do
+      assert :erlang.binary_part("goldfish", 0, 4) == "gold"
+    end
+
+    test "start is at the end of the binary" do
+      assert :erlang.binary_part("goldfish", 8, -4) == "fish"
+    end
+
+    test "third argument is negative" do
+      assert :erlang.binary_part("golden retriever", 16, -9) == "retriever"
+    end
+
+    test "third argument is zero" do
+      assert :erlang.binary_part("goldfish", 2, 0) == ""
+    end
+
+    test "from the middle of the binary" do
+      assert :erlang.binary_part("golden retriever", 7, 9) == "retriever"
+    end
+
+    test "takes whole binary" do
+      assert :erlang.binary_part("goldfish", 0, 8) == "goldfish"
+    end
+
+    test "takes whole binary reversed" do
+      assert :erlang.binary_part("goldfish", 8, -8) == "goldfish"
+    end
+
+    test "raises ArgumentError if the first argument is not a binary" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "not a binary"),
+                   {:erlang, :binary_part, [1, 1, 1]}
+    end
+
+    test "raises ArgumentError if the first argument is a non-binary bitstring" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "not a binary"),
+                   {:erlang, :binary_part, [<<1::1, 0::1, 1::1>>, 1, 1]}
+    end
+
+    test "raises ArgumentError if the second argument is not an integer" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(2, "not an integer"),
+                   {:erlang, :binary_part, ["goldfish", true, 2]}
+    end
+
+    test "raises ArgumentError if the third argument is not an integer" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(3, "not an integer"),
+                   {:erlang, :binary_part, ["goldfish", 1, true]}
+    end
+
+    test "raises ArgumentError if the second argument is negative" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(2, "out of range"),
+                   {:erlang, :binary_part, ["goldfish", -1, 2]}
+    end
+
+    test "raises ArgumentError if the second argument is larger than the size of the first argument" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(2, "out of range"),
+                   {:erlang, :binary_part, ["goldfish", 9, 2]}
+    end
+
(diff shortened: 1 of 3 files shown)


test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "binary_part/3" do
```
bartblast, line 1376: Could you move the error case tests to the end (successful cases first)?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+    test "raises ArgumentError if the first argument is a non-binary bitstring" do
```
bartblast, line 1428: Could I add a test case for when the first arg is not a bitstring?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "not a binary"),
+                   {:erlang, :binary_part, [<<1::1, 0::1, 1::1>>, 1, 1]}
+    end
+
+    test "raises ArgumentError if the second argument is a boolean" do
```
bartblast: I assume you're testing a case where the second argument is not an integer, correct? If so, could you update the description to reflect this intent?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+      assert_error ArgumentError,
+                   build_argument_error_msg(2, "not an integer"),
+                   {:erlang, :binary_part, ["goldfish", true, 2]}
+    end
+
+    test "raises ArgumentError if the third argument is a boolean" do
```
bartblast: I assume you're testing a case where the third argument is not an integer, correct? If so, could you update the description to reflect this intent?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+                   {:erlang, :binary_part, ["goldfish", 0, 9]}
+    end
+
+    test "raises ArgumentError if the third argument is negative and the second and third arguments summed is smaller than zero" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(3, "out of range"),
+                   {:erlang, :binary_part, ["goldfish", 4, -5]}
+    end
+
+    test "succeeds if the first argument is a text binary" do
```
bartblast: "succeeds if (..." feels redundant IMO. Could you remove it here and in other tests?
Something like this should be enough: "first argument is a text binary"

test/javascript/erlang/erlang_test.mjs
```diff
@@ -1531,6 +1531,195 @@ describe("Erlang", () => {
     });
   });
 
+  describe("binary_part/3", () => {
```
bartblast, line 1626: Comments for `erlang_test.exs` apply here as well.

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+
+    test "succeeds if the second argument is zero" do
+      assert :erlang.binary_part("goldfish", 0, 4) == "gold"
+    end
+
+    test "succeeds if the second argument is the size of the first argument" do
+      assert :erlang.binary_part("goldfish", 8, -4) == "fish"
+    end
+
+    test "succeeds if the third argument is negative" do
```
bartblast: Maybe it makes sense to make the tests slightly different than the previous one?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+    test "succeeds if the first argument is a byte binary" do
+      assert :erlang.binary_part(<<1, 2, 3, 4, 5, 6, 7, 8, 9, 10>>, 4, 2) == <<5, 6>>
+    end
+
+    test "succeeds if the first argument is empty" do
+      assert :erlang.binary_part("", 0, 0) == ""
+    end
+
+    test "succeeds if the first argument contains multi-byte unicode characters" do
+      assert :erlang.binary_part("á", 0, 1) == <<195>>
```
bartblast: Could you add the same comment as in the corresponding JS test?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+    end
+
+    test "succeeds if the second argument is zero" do
+      assert :erlang.binary_part("goldfish", 0, 4) == "gold"
+    end
+
+    test "succeeds if the second argument is the size of the first argument" do
```
bartblast: "start is at the end of the binary" (or something like that) ?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
bartblast: Maybe it makes sense to make the tests slightly different than the previous one?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1317,6 +1317,96 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+
+    test "succeeds if the third argument is zero" do
+      assert :erlang.binary_part("goldfish", 2, 0) == ""
+    end
+
+    test "succeeds from the middle of the binary" do
+      assert :erlang.binary_part("goldfish", 2, 4) == "ldfi"
+    end
+
+    test "succeeds with the whole binary" do
```
bartblast: Another similar case for the whole binary but starting from the end would be useful.

bartblast:

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
djonn: Do you generally prefer naming the arguments (binary/subject, start, length) or counting them (first argument)?

djonn:
I have rebased onto dev and addressed your comments

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
bartblast: It depends. When we're testing argument type validation, it makes sense to use ordinals like "first argument", "second argument", etc. In other cases, it usually makes sense to use the parameter name or describe the intent of the test instead of just using parameter names or ordinals.

bartblast:
Could you resolve branch conflicts?

bartblast:
Looks good! Thanks @djonn! 💜

PR #323, opened 2025-11-16
Port :binary.copy/2 to JS

hlorellium:


assets/js/erlang/binary.mjs
```diff
@@ -0,0 +1,58 @@
...
+            return Bitstring.fromText("");
+        }
+
+        const count = Number(n.value);
+
+        if (subject.text !== null) {
+            return Bitstring.fromText(subject.text.repeat(count));
+        }
+
+        Bitstring.maybeSetBytesFromText(subject);
```
bartblast: This is redundant, if `subject.text` is null (preceding check), then `subject.bytes` is not null.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+  Each Elixir consistency test has a related JavaScript test in test/javascript/erlang/binary_test.mjs
+  Always update both together.
+  """
+
+  use Hologram.Test.BasicCase, async: true
+
+  @moduletag :consistency
+
+  describe "copy/2" do
+    test "copies a text-based binary multiple times" do
```
bartblast: Could you make the success case tests and test names more consistent?
We could have a text "matrix" with 3 dimensions:
1) empty / non-empty
2) text-based / bytes-based
3) zero times / single time / multiple times
and then create tests according to this pattern:
"copies a non-empty, text-based binary multiple times", and so on...
That makes 12 success case tests in total.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      assert :binary.copy("", 5) === ""
+    end
+
+    test "copies a byte-based binary" do
+      binary = <<65, 66, 67>>
+      result = :binary.copy(binary, 2)
+      assert result === <<65, 66, 67, 65, 66, 67>>
+    end
+
+    test "copies a binary with a large multiplier" do
```
bartblast: This test is redundant IMO - the multiplier is arbitrary and doesn't verify anything beyond what a smaller value would test.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      result = :binary.copy("x", 1000)
+      assert result === String.duplicate("x", 1000)
+      assert byte_size(result) === 1000
+    end
+
+    test "raises ArgumentError if the first argument is not a binary (atom)" do
+      assert_error(
+        ArgumentError,
+        "errors were found at the given arguments:\n\n  * 1st argument: not a binary\n",
```
bartblast: Could you use `build_argument_error_msg/2` helper here and in other similar tests?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    test "copies a binary with a large multiplier" do
+      result = :binary.copy("x", 1000)
+      assert result === String.duplicate("x", 1000)
+      assert byte_size(result) === 1000
+    end
+
+    test "raises ArgumentError if the first argument is not a binary (atom)" do
+      assert_error(
```
bartblast: Could you remove parenthesis (here and in other tests)?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      assert result === <<65, 66, 67, 65, 66, 67>>
+    end
+
+    test "copies a binary with a large multiplier" do
+      result = :binary.copy("x", 1000)
+      assert result === String.duplicate("x", 1000)
+      assert byte_size(result) === 1000
+    end
+
+    test "raises ArgumentError if the first argument is not a binary (atom)" do
```
bartblast: Could you change it to: "raises ArgumentError if the first argument is not a bitstring" ?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      assert_error(
+        ArgumentError,
+        "errors were found at the given arguments:\n\n  * 1st argument: not a binary\n",
+        {:binary, :copy, [:not_binary, 2]}
+      )
+    end
+
+    test "raises ArgumentError if the first argument is not a binary (integer)" do
```
bartblast: This test is redundant.

Could you test a non-binary bitstring instead?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+    test "raises ArgumentError if the second argument is not an integer" do
+      assert_error(
+        ArgumentError,
+        "errors were found at the given arguments:\n\n  * 2nd argument: not an integer\n",
+        {:binary, :copy, ["test", :not_integer]}
+      )
+    end
+
+    test "raises ArgumentError if n is negative" do
```
bartblast: n -> second argument (or count)

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      assert_error(
+        ArgumentError,
+        "errors were found at the given arguments:\n\n  * 2nd argument: out of range\n",
+        {:binary, :copy, ["test", -1]}
+      )
+    end
+
+    test "raises ArgumentError if the first argument is a non-byte-aligned bitstring" do
```
bartblast: This is redundant. We need a test case for a non-binary bitstring, though. See one of the preceding comments.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      bitstring = <<1::3>>
+
+      assert_error(
+        ArgumentError,
+        "errors were found at the given arguments:\n\n  * 1st argument: is a bitstring (expected a binary)\n",
+        {:binary, :copy, [bitstring, 2]}
+      )
+    end
+
+    test "always returns a new binary (even with n=1)" do
```
bartblast: Generally your idea makes sense here, but it would have to be applied across the codebase and I decided not to do it for various reasons in the past.

assets/js/erlang/binary.mjs
```diff
@@ -0,0 +1,58 @@
...
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const Erlang_Binary = {
+    // Start copy/2
+    "copy/2": function (subject, n) {
```
bartblast: I know Erlang uses `n` for this param, but can you use `count` instead? Erlang params are often cryptic... 🤷‍♂️

test/javascript/erlang/binary_test.mjs
```diff
@@ -0,0 +1,159 @@
...
+describe("Erlang_Binary", () => {
+  describe("copy/2", () => {
+    const testedFun = Erlang_Binary["copy/2"];
+
+    it("copies a text-based binary multiple times", () => {
+      const binary = Bitstring.fromText("hello");
+      const n = Type.integer(3);
+      const result = testedFun(binary, n);
+
+      assert.strictEqual(result.text, "hellohellohello");
```
bartblast: Could you use `assertBoxedStrictEqual()` for comparisons (here and in other tests)? This way you will be able to compare boxed value objects instead of testing fields.

test/javascript/erlang/binary_test.mjs
```diff
@@ -0,0 +1,159 @@
+"use strict";
```
bartblast, line 1: The comments for `binary_test.exs` also apply here.

bartblast:
Besides the review I've just submitted, could you also:
1) run `$ mix f` in the root dir to format code (see the CI checks)
2) sync with dev, since there was another PR related to :binary module merged to dev

bartblast:

bartblast:
Thank you @hlorellium! :)

PR #334, opened 2025-11-17
Port :filename.basename/1 to JS

danirukun:
Closes https://github.com/bartblast/hologram/issues/324

### DIFF
--- a/assets/js/erlang/filename.mjs
+++ b/assets/js/erlang/filename.mjs
@@ -28,9 +28,7 @@ const Erlang_Filename = {
       filepathText = binary.text;
       returnAsCodepoints = true;
     } else if (Type.isAtom(filename)) {
-      const binary = Erlang["atom_to_binary/1"](filename);
-      Bitstring.maybeSetTextFromBytes(binary);
-      filepathText = binary.text;
+      filepathText = filename.value;
       returnAsCodepoints = true;
     } else {
       Interpreter.raiseFunctionClauseError(
@@ -50,7 +48,7 @@ const Erlang_Filename = {
       : basenameBinary;
   },
   // End basename/1
-  // Deps: [:erlang.atom_to_binary/1, :erlang.iolist_to_binary/1]
+  // Deps: [:erlang.iolist_to_binary/1]
 };
 
 export default Erlang_Filename;
--- a/lib/hologram/compiler/call_graph.ex
+++ b/lib/hologram/compiler/call_graph.ex
@@ -31,7 +31,6 @@ defmodule Hologram.Compiler.CallGraph do
     {{:erlang, :integer_to_binary, 1}, {:erlang, :integer_to_binary, 2}},
     {{:erlang, :iolist_to_binary, 1}, {:lists, :flatten, 1}},
     {{:erlang, :split_binary, 2}, {:erlang, :byte_size, 1}},
-    {{:filename, :basename, 1}, {:erlang, :atom_to_binary, 1}},
     {{:filename, :basename, 1}, {:erlang, :iolist_to_binary, 1}},
     {{:lists, :keymember, 3}, {:lists, :keyfind, 3}},
     {{:maps, :get, 2}, {:maps, :get, 3}},


bartblast:

assets/js/erlang/filename.mjs
```diff
@@ -0,0 +1,67 @@
...
+  "basename/1": (atomOrBinaryOrList) => {
+    if (Type.isList(atomOrBinaryOrList)) {
+      if (atomOrBinaryOrList.data.length === 0) {
+        return Type.list([]);
+      }
+
+      const binary = Erlang["iolist_to_binary/1"](atomOrBinaryOrList);
+      Bitstring.maybeSetTextFromBytes(binary);
+      const parts = binary.text.split("/").filter((part) => part !== "");
+      const basenameText = parts.length > 0 ? parts[parts.length - 1] : "";
```
bartblast: `parts[parts.length - 1]`
->
`parts.at(-1)`

assets/js/erlang/filename.mjs
```diff
@@ -0,0 +1,67 @@
...
+      const basenameBinary = Type.bitstring(basenameText);
+      return Bitstring.toCodepoints(basenameBinary);
+    }
+
+    if (
```
bartblast: !Type.isBinary() alone is enough

assets/js/erlang/filename.mjs
```diff
@@ -0,0 +1,67 @@
+"use strict";
+
+import Bitstring from "../bitstring.mjs";
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+
+const Erlang_Filename = {
+  // Start basename/1
+  "basename/1": (atomOrBinaryOrList) => {
```
bartblast: IMO you can substantially simplify the code if you write it this way:

```js
let filepathText;

if (Type.isBinary(filepath)) { // this will be the most common case
   // binary-specific behaviour and conversion
  ...
  filepathText = ...;
} else if (Type.isList(term) {
  ...
} else if (Type.isAtom(term) {
  ...
} else {
  // here we're raising the exception
}

const parts = filepathText.split("/").filter((part) => part !== "");
...
```

bartblast: Could you use `filename` instead of `atomOrBinaryOrList`?

assets/js/erlang/filename.mjs
```diff
@@ -0,0 +1,67 @@
+"use strict";
+
+import Bitstring from "../bitstring.mjs";
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+
```
bartblast, line 11: Please keep the preamble:

```
// IMPORTANT!
// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
```

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
+defmodule Hologram.ExJsConsistency.Erlang.FilenameTest do
+  @moduledoc """
+  IMPORTANT!
+  Each Elixir consistency test has a related JavaScript test in test/javascript/erlang/filename_test.mjs
+  Always update both together.
+  """
```
bartblast, line 6: Add newline between, please.

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
...
+  use Hologram.Test.BasicCase, async: true
+
+  @moduletag :consistency
+
+  describe "basename/1" do
+    test "returns basename for path with multiple segments" do
```
bartblast: "returns basename for " prefix for all of these tests is redundant IMO,
"path with multiple segments" is enough.

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
...
+
+    test "returns basename for path with unicode characters" do
+      assert :filename.basename("path/to/文件.txt") == "文件.txt"
+    end
+
+    test "returns basename for path with special characters" do
+      assert :filename.basename("path/to/file-name_123.txt") == "file-name_123.txt"
+    end
+
+    test "returns basename as list of code points for atom input" do
```
bartblast: One test for non-nil value and one for nil is enough I think...

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
...
+
+    test "returns basename as list of code points for nil atom input with path" do
+      assert :filename.basename(:"path/to/nil") == ~c"nil"
+    end
+
+    test "returns basename as list of code points for atom input with unicode characters" do
+      assert :filename.basename(:"path/to/文件.txt") == ~c"文件.txt"
+    end
+
+    test "returns empty list for empty list input" do
```
bartblast, line 57: Similar to atoms, testing one empty list and one non-empty list is sufficient. Otherwise, we're essentially testing `:erlang.atom_to_binary/1` and `:erlang.iolist_to_binary/1` themselves.
(This applies if you simplify the function as I outlined initially.)

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
...
+
+    test "returns basename as list of code points for non-empty list input" do
+      assert :filename.basename([~c"path/to/", ?f, ?i, ?l, ?e, ~c".txt"]) == ~c"file.txt"
+    end
+
+    test "returns basename as list of code points for list input with single segment" do
+      assert :filename.basename([?f, ?i, ?l, ?e, ?n, ?a, ?m, ?e]) == ~c"filename"
+    end
+
+    test "raises FunctionClauseError if the argument is not a binary or atom" do
```
bartblast: ... or list

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
...
+      assert_error FunctionClauseError,
+                   build_function_clause_error_msg(":filename.do_flatten/2", [123, []]),
+                   fn -> :filename.basename(123) end
+    end
+
+    test "raises FunctionClauseError if the argument is a tuple" do
```
bartblast: This is redundant, otherwise we'd need to test all other types.

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -0,0 +1,109 @@
...
+      assert_error FunctionClauseError,
+                   build_function_clause_error_msg(":filename.do_flatten/2", [{1, 2}, []]),
+                   fn -> :filename.basename({1, 2}) end
+    end
+
+    test "raises FunctionClauseError if the argument is a non-binary bitstring" do
```
bartblast, line 75: However, this one is worth keeping.

test/javascript/erlang/filename_test.mjs
```diff
@@ -0,0 +1,225 @@
...
+      const result = basename(Type.bitstring("path/to/file-name_123.txt"));
+      const expected = Type.bitstring("file-name_123.txt");
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it("returns basename as list of code points for atom input", () => {
+      const result = basename(Type.atom("path/to/file.txt"));
+      const expected = Type.list(
+        Array.from("file.txt", (char) => Type.integer(char.codePointAt(0))),
```
bartblast: Could you create a test helper `iolist(string)` in `helpers.mjs` and apply it in other `basename/1` tests?

test/javascript/erlang/filename_test.mjs
```diff
@@ -0,0 +1,225 @@
+"use strict";
```
bartblast, line 1: The comments from Elixir consistency tests apply here too.

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
danirukun: You mean for bitstring and atoms? Do we only drop the unit tests that test the deeper conversion mechanics from the Erlang module, e.g. Unicode conversions etc.?

bartblast: We want to test only the logic of `:filename.basename/1`, not deeply test the underlying conversion functions (those should have their own tests).

Our testing strategy should be:

1. **Exercise each code path** in the basename function - we need code coverage of all the logical branches (empty path, single segment, multiple segments, trailing slashes, etc.)

2. **Test each allowed input type** - we need to exercise the code branches for each type (atom, bitstring, etc.). Additionally, since the code uses `maybeSetTextFromBytes()`, we need to test both **text-based and bytes-based input for each type** (I've just noticed this requirement). This ensures both internal representations are handled correctly.

3. **Test error handling** with invalid input types

So the issue with having both:
- "list of code points for atom input"
- "list of code points for atom input with single segment"

...is that they're testing the **same code path** (non-empty atom input) twice. We should test each distinct code path once per type-representation combination, but we don't need to test every behavioral scenario (empty path, single segment, multiple segments, etc.) for every type.

**In practice this means:**
- Core behavioral tests: empty path, single segment, multi-segment, trailing slash, etc. (exercise all code paths)
- For each allowed input type: test both text-based AND bytes-based representations if there may be conversion with `maybeSetTextFromBytes()`
- Invalid input error handling

This gives us full code coverage without combinatorial explosion. Ideally, we'd test all combinations of (behavioral scenarios × input types × representations), but in this function that would mean combinatorial explosion, making the tests unmaintainable.

assets/js/erlang/filename.mjs
```diff
@@ -0,0 +1,56 @@
...
+      }
+
+      const binary = Erlang["iolist_to_binary/1"](filename);
+      Bitstring.maybeSetTextFromBytes(binary);
+      filepathText = binary.text;
+      returnAsCodepoints = true;
+    } else if (Type.isAtom(filename)) {
+      const binary = Erlang["atom_to_binary/1"](filename);
+      Bitstring.maybeSetTextFromBytes(binary);
+      filepathText = binary.text;
```
bartblast: Here, this can be simplified to:
`filepathText = filename.text`

test/javascript/support/helpers.mjs
```diff
@@ -334,3 +334,9 @@ export function registerWebApis() {
 export function waitForEventLoop() {
   return new Promise((resolve) => setTimeout(resolve, 0));
 }
+
+export function iolist(string) {
```
bartblast: Could you order the functions alphabetically?

assets/js/erlang/filename.mjs
danirukun: Hmm are you sure? If I do that it ends up being undefined

bartblast: Sorry, I meant `filename.value`.

danirukun: ah thought so, gotcha

danirukun:
If you want before merge I can clean up the commits and squash them a bit.

bartblast:
No need, I squash them during the merge anyway. Thanks for your help @danirukun! :)

PR #343, opened 2025-11-17
Port :math.exp/1 function to JS

Hyftar:
Closes #341 

## Summary by CodeRabbit

* **New Features**
  * Introduced `exp/1` function to calculate exponential values, supporting all numeric types with automatic conversion
  * Introduced `pow/2` function for power operations with base and exponent parameters, featuring full numeric type validation and error handling for special cases including fractional exponents applied to negative bases

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #347, opened 2025-11-17
Port :math.ceil/1 to JS

### LINKED ISSUE #345: Port :math.ceil/1 to JS
I'm not sure if this needs to be ported as well since we have #332 already.

Hyftar:
Closes #345


Hyftar:
The CI is failing but I don't think this is related to these changes

bartblast:

bartblast:
Hi @Hyftar,

> The CI is failing but I don't think this is related to these changes

The CI is failing because your tests emit warnings, see:
```
warning: the call to :math.ceil/1 will fail with ArgumentError
Warning:   test/elixir/hologram/ex_js_consistency/erlang/math_test.exs:44

warning: the call to :math.ceil/1 will fail with ArgumentError
Warning:   test/elixir/hologram/ex_js_consistency/erlang/math_test.exs:52

warning: the call to :math.exp/1 will fail with ArgumentError
Warning:   test/elixir/hologram/ex_js_consistency/erlang/math_test.exs:98

warning: the call to :math.exp/1 will fail with ArgumentError
Warning:   test/elixir/hologram/ex_js_consistency/erlang/math_test.exs:106
```
Elixir compiler detects that the calls you're testing will fail anyway. In such cases we use different syntax - instead of running anonymous function we use something like this (taken from another function):
```elixir
assert_error ArgumentError,
  build_argument_error_msg(1, "not a bitstring"),
  {:erlang, :byte_size, [:abc]}
```
  
---

Also could you implement more robust tests? You can base the tests on `:erlang.ceil/1` tests:
https://github.com/bartblast/hologram/blob/bd76e6964b6e09d7c71175874043026dd2092382/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs#L1502
(if you copy those tests it will also fix your CI warnings)

---

> ⚠️ This PR depends on #343; Please merge it first!

Could you remove this function from the other PR and keep one function per PR (as per the Contributing Guide)?

Hyftar:
> Could you remove this function from the other PR and keep one function per PR (as per the Contributing Guide)?

I think you misunderstood what I meant by that; I meant the other PR should be merged first since this one depends on it.

I'm not sure how you want me to proceed otherwise, since the `math` module did not exist before the other PR, should I just recreate it in both PRs? This seems rather inefficient to me; which is why I based this branch on my other branch. Once the first PR is merged, there will be no conflicts with this one, but if both re-create the same module, there will be conflicts.

bartblast:
Hi @Hyftar,

I understand exactly what you mean - PR #347 "depends" on #343 because you branched it from there. I see the situation clearly.

However, this is a **self-imposed dependency**, not a technical one. The functions don't call each other - they're independent. You could have created two separate branches from `dev`, each with its own `math.mjs` file containing one function. That's the workflow we need contributors to follow.

**One function per PR** is essential because:

1. **Coordination** - Multiple contributors may be working on the same module simultaneously. Separate PRs prevent blocking and merge conflicts.

2. **Isolated CI checks** - We need to test each function in isolation to catch function-specific issues without noise from other changes.

3. **Faster reviews** - Small, focused PRs get reviewed and merged much quicker than bundled ones.

You can still fix this: create two separate branches from `dev` in your fork, each with its own copy of the module file. When the first PR merges, you'll just need to resolve a simple conflict in the second one.

I know it might seem less efficient from your perspective, but this workflow makes my life (and the project's progress) significantly easier.

Could you please split these into separate PRs? Thanks for understanding!

bartblast:
Hi @Hyftar,

I understand exactly what you mean - PR #347 depends on #343 because you branched it from there. I see the situation clearly.

However, this is a **self-imposed dependency**, not a technical one. The functions don't call each other - they're independent. You could have created two separate branches from `dev`, each with its own `math.mjs` file containing one function. That's the workflow we need contributors to follow.

**One function per PR** is essential because:

1. **Coordination** - Multiple contributors can work on the same module in parallel without blocking each other. When conflicts arise, they're simple to resolve since functions are independent.

2. **Isolated CI checks** - We need to test each function in isolation to catch function-specific issues without noise from other changes.

3. **Faster reviews** - Small, focused PRs get reviewed and merged much quicker than bundled ones.

Here's how to fix this: If PR #343 contains only `:math.exp/1`, it can stay as is. For `:math.ceil/1`, you'll need to create a new PR from either:
- A new branch from `dev` (not from your #343 branch) with only `:math.ceil/1`, or
- A separate fork if you prefer complete isolation

Either way, both PRs will temporarily create/modify the same file, but that's fine - the second one will just need a simple merge conflict resolution.

I know it requires rework, but this workflow makes my life (and the project's progress) significantly easier.

Thanks for understanding!

Hyftar:
@bartblast, no worries, I just thought having no merge conflicts would have been easier on you, but if you'd rather have merge conflicts than dependant PRs, I can work around that.

I'll rebase my PR to remove the dependant commits.

assets/js/erlang/math.mjs
```diff
@@ -0,0 +1,26 @@
...
+import Type from "../type.mjs";
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const Erlang_Math = {
+  // Start ceil/1
+  "ceil/1": (x) => {
+    if (!Type.isFloat(x) && !Type.isInteger(x)) {
```
bartblast: Could you change it `!Type.isNumber()` ?

assets/js/erlang/math.mjs
```diff
@@ -0,0 +1,26 @@
...
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const Erlang_Math = {
+  // Start ceil/1
+  "ceil/1": (x) => {
```
bartblast: Could you change x -> number ?

assets/js/erlang/math.mjs
```diff
@@ -0,0 +1,26 @@
...
+    if (!Type.isFloat(x) && !Type.isInteger(x)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(1, "not a number")
+      );
+    }
+
+    const numericValue = Type.isInteger(x) ? Number(x.value) : x.value;
+    return Type.float(Math.ceil(numericValue));
```
bartblast: This should be slightly simpler and faster (no need to ceil integers):
`return Type.isInteger(x) ? Type.float(Number(x.value)) : Type.float(Math.ceil(x.value));`

assets/eslint.config.mjs
```diff
@@ -32,6 +32,7 @@ export default [
         Erlang: "readonly",
         Erlang_Code: "readonly",
         Erlang_Lists: "readonly",
+        Erlang_Math: "readonly",
```
bartblast: We need alphabetical ordering here.

bartblast:
Looks good, thanks for help @Hyftar! 💜

PR #350, opened 2025-11-17
Port :binary.first/1 to JS

### LINKED ISSUE #320: Port :binary.first/1 to JS
Returns the first byte of binary Subject as an integer.

If the size of Subject is zero, a badarg exception is raised.

## Examples

    1> binary:first(<<42,99,100>>).
    42
    2> binary:first(<<>>).
    ** exception error: bad argument
         in function  binary:first/1
            called as binary:first(<<>>)
            *** argument 1: a zero-sized binary is not allowed

franzinBr:
Closes #320 

## Summary by CodeRabbit

* **New Features**
  * Added `binary.first/1` function to retrieve the first byte value from a binary, returning it as an integer.

* **Tests**
  * Added comprehensive test coverage for `binary.first/1`, validating correct behavior with single-byte and multi-byte binaries, text strings, UTF-8 characters, and error handling for invalid inputs.

### DIFF
--- a/assets/js/erlang/binary.mjs
+++ b/assets/js/erlang/binary.mjs
@@ -68,6 +68,8 @@ const Erlang_Binary = {
     return Type.integer(subject.bytes[0]);
   },
   // End first/1
+  // Deps: []
+
   // Start copy/2
   "copy/2": (subject, count) => {
     if (!Type.isBinary(subject)) {


bartblast:

assets/js/erlang/binary.mjs
```diff
@@ -41,6 +41,39 @@ const Erlang_Binary = {
   },
   // End at/2
   // Deps: []
+
+  // Start first/1
+  "first/1": (subject) => {
+    if (!Type.isBinary(subject) && Type.isBitstring(subject)) {
```
bartblast: You can simplify this and decrease the number of checks with:

```js
if (!Type.isBinary(subject)) {
  let message = Type.isBitstring(subject) ? "is a bitstring (expected a binary)" : "not a binary";

    Interpreter.buildArgumentErrorMsg(1, message),
  );
}
```

assets/js/erlang/binary.mjs
```diff
@@ -41,6 +41,39 @@ const Erlang_Binary = {
...
+      );
+    }
+
+    if (!Type.isBinary(subject)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(1, "not a binary"),
+      );
+    }
+
+    Bitstring.maybeSetBytesFromText(subject);
```
bartblast: This line should go after the `isEmpty()` check block.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -56,4 +56,97 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
                    fn -> :binary.at(@binary, -1) end
     end
   end
+
+  describe "first/1" do
+    test "raises ArgumentError if the first argument is integer" do
```
bartblast: Could you move the error case tests to the end and use "subject" instead of "first argument"?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -56,4 +56,97 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+                   {:binary, :first, [{1, 2}]}
+    end
+
+    test "raises ArgumentError if the first argument is a map" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "not a binary"),
+                   {:binary, :first, [%{key: :value}]}
+    end
+
+    test "raises ArgumentError if the first argument is a non-binary bitstringg" do
```
bartblast: Found a typo: "bitstringg"

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -56,4 +56,97 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+                   {:binary, :first, [<<>>]}
+    end
+
+    test "raises ArgumentError if the first argument is a zero-sized text" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "a zero-sized binary is not allowed"),
+                   {:binary, :first, [""]}
+    end
+
+    test "return first byte of a binary" do
```
bartblast: You just only need a single test that tests text version vs bytes version, to make sure that the bytes conversion is corrrectly applied. Otherwise you're essentially testing the `<< >>` operator or conversion.

And then for testing specifying behaviours you can just test on the text OR bytes version, whatever you fell is more convenient.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -56,4 +56,97 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+    test "returns first byte of a single-character text" do
+      assert :binary.first("Z") == 90
+    end
+
+    test "returns first byte of a large binary" do
+      large_binary = <<123>> <> :binary.copy(<<0>>, 999)
+      assert :binary.first(large_binary) == 123
+    end
+
+    test "returns first byte of binary created from float literal" do
```
bartblast: Redundant, this tests `<< >>` operator.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -56,4 +56,97 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
                    fn -> :binary.at(@binary, -1) end
     end
   end
+
+  describe "first/1" do
```
bartblast, line 60: Most of the tests are actually redundant.

1) You don't need to test each behaviour on text-based vs bytes-based version, you just need to test that the conversion to bytes is correctly applied in one case, and then use either bytes or text depending on what is more convenient for the given case.  Otherwise you're essentially testing the `<< >>` operator or internal conversion.

2) When testing type validation, we just need:
* binary bitstring
* non-binary bitstring
* not bitstring

3) Another testing dimension:
* zero bytes
* single byte
* multiple bytes

and that should be enough IMO.

test/javascript/erlang/binary_test.mjs
```diff
@@ -87,4 +87,158 @@ describe("Erlang_Binary", () => {
       );
     });
   });
+
+  describe("first/1", () => {
```
bartblast, line 105: Comments for the Elixir consistency tests apply here too.

bartblast (changes requested):

bartblast:
@coderabbitai full review

bartblast:
Looks good, thank you @franzinBr! 💜

PR #331, opened 2025-11-16
Port :erlang.tuple_size/1 to JS

brGuirra:
Closes #330

PR #360, opened 2025-11-18
Port :erlang.setelement/3 to JS

### LINKED ISSUE #359: Port :erlang.setelement/3 to JS

Sorc96:
Closes #359

Currently checks the length of the tuple's data directly to determine whether the index is in or out of range. Could be changed to use `:erlang.tuple_size/1` after #331 is merged.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
@@ -2012,7 +2012,15 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
 
   describe "setelement/3" do
     test "replaces the element at the given index" do
+      assert :erlang.setelement(2, {1, 2, 3}, :a) === {1, :a, 3}
+    end
+
+    test "replaces the first element" do
       assert :erlang.setelement(1, {1, 2}, :a) === {:a, 2}
+    end
+
+    test "replaces the last element" do
+      assert :erlang.setelement(2, {1, 2}, :a) === {1, :a}
     end
 
     test "raises ArgumentError if the first argument is not an integer" do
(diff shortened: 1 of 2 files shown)


bartblast:

assets/js/erlang/erlang.mjs
```diff
@@ -881,6 +881,35 @@ const Erlang = {
...
+      );
+    }
+
+    if (!Type.isTuple(tuple)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(2, "not a tuple"),
+      );
+    }
+
+    if (index.value <= 0 || index.value > tuple.data.length) {
```
bartblast: `index.value <= 0n`
will be marginally faster

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -2010,6 +2010,36 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "setelement/3" do
+    test "replaces the element at the given index" do
```
bartblast, line 2014: Could you test boundary conditions?
* first item
* middle item
* last item

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2767,6 +2767,47 @@ describe("Erlang", () => {
     });
   });
 
+  describe("setelement/3", () => {
+    const setelement = Erlang["setelement/3"];
+
+    it("replaces the element at the given index", () => {
```
bartblast, line 2773: Could you test boundary conditions?
* first item
* middle item
* last item

bartblast:
> Currently checks the length of the tuple's data directly to determine whether the index is in or out of range. Could be changed to use `:erlang.tuple_size/1` after #331 is merged.

I'm ok to do this directly in this function as a performance optimization (one less input validation - since the other function validates its input as well).

assets/js/erlang/erlang.mjs
Sorc96: Thanks, that was an oversight on my part.

bartblast:
Thanks for the updates! :)

PR #370, opened 2025-11-18
WIP Port :erlang.float_to_list/2 to JS

### LINKED ISSUE #354: Port :erlang.float_to_list/2 to JS
https://hologram.page/reference/client-runtime/erlang/erlang/float_to_list/2

return a list of Codepoints or a string?

ideaMarcos:
Closes #354

### DIFF
--- a/assets/js/erlang/erlang.mjs
+++ b/assets/js/erlang/erlang.mjs
@@ -359,6 +359,32 @@ const Erlang = {
   // End binary_to_existing_atom/2
   // Deps: [:erlang.binary_to_atom/2]
 
+  // Start binary_to_float/1
+  "binary_to_float/1": (binary) => {
+    if (!Type.isBinary(binary)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(1, "not a binary"),
+      );
+    }
+
+    const text = Bitstring.toText(binary);
+
+    const floatRegex = /^[+-]?\d+\.\d+([eE][+-]?\d+)?$/;
+
+    if (!floatRegex.test(text)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(
+          1,
+          "not a textual representation of a float",
+        ),
+      );
+    }
+
+    return Type.float(Number(text));
+  },
+  // End binary_to_float/1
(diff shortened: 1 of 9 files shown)


assets/js/erlang/erlang.mjs
```diff
@@ -547,6 +547,42 @@ const Erlang = {
   // End float_to_binary/2
   // Deps: []
 
+  // Start float_to_list/2
+  "float_to_list/2": (float, opts) => {
+    if (!Type.isFloat(float)) {
```
ideaMarcos: this logic is similar to float_to_binary/2 above. Should this call float_to_binary/2 like so?

```elixir
float_to_binary(float, opts)
|> Bitstring.toCodepoints()
```

bartblast: Yeah, definitely! Wanna help on the opts param for float_to_binary/2 (in a separate issue)?

ideaMarcos: sure i'll give it a shot

bartblast: @ideaMarcos,

Then, I think the best course of action is:
1. Create a separate issue for implementing the opts argument in :erlang.float_to_binary/2 and implement it (including tests for the options)
2. Once the pull request for point 1 is merged, pull the changes from dev and finalize this pull request

Does that make sense?

ideaMarcos: sure.  you can create one ticket or separate tickets for decimals, compact, and scientific options. up to you

bartblast: Cool, here you go: https://github.com/bartblast/hologram/issues/371

bartblast: (could you make any comment in the issue, I can't assign you otherwise)

bartblast:
To format code you can use:
`$ mix f`
in the root dir

bartblast:

bartblast:
@ideaMarcos, could you pull the new changes from `dev` branch?

 After pulling, could you remove all the current tests and replace them with a single test that verifies the function delegates to `:float_to_binary/2` and converts the result to codepoints? For the test case, you could use an example call with a non-default option (this way we'll be sure that the options are passed correctly).

PR #385, opened 2025-11-20
Support the rest of options in :erlang.float_to_binary/2

### LINKED ISSUE #371: Support the rest of options in :erlang.float_to_binary/2
Option `:short` is already supported.

**Options to support:**

`:decimals`
If option decimals is specified, the returned value contains at most Decimals number of digits past the decimal point. If the number does not fit in the internal static buffer of 256 bytes, the function throws badarg.

`:compact`
If option compact is specified, the trailing zeros at the end of the list are truncated. This option is only meaningful together with option decimals.

`:scientific`
If option scientific is specified, the float is formatted using scientific notation with Decimals digits of precision.

See section "Available options" in `:erlang.float_to_list/2` docs (the behaviour is the same for `:erlang.float_to_binary/2`):
https://www.erlang.org/doc/apps/erts/erlang.html#float_to_list/2

ideaMarcos:
Closes #371


assets/js/erlang/erlang.mjs
```diff
@@ -532,17 +532,61 @@ const Erlang = {
...
+        );
+      }
+    }
+
+    let result;
+
+    if (isShort) {
+      result = float.value.toString();
+    } else if (scientific !== null) {
+      result = float.value.toExponential(scientific);
```
ideaMarcos: exponent format is different in erlang/js

bartblast: Can you explain the differences?

ideaMarcos: sorry, i left comment in code

Erlang format uses zero-padded exponents (e.g., e+00, e-04)
```elixir
iex> :erlang.float_to_binary(712.0, [{:scientific, 3}])
"7.120e+02"
```

JavaScript may use e+0, e-4, so we need to pad the exponent
```javascript
> console.log(Number.parseFloat(712.0).toExponential(3));
"7.120e+2"
```

assets/js/erlang/erlang.mjs
```diff
@@ -589,17 +589,60 @@ const Erlang = {
...
-    ) {
-      throw new HologramInterpreterError(
-        ":erlang.float_to_binary/2 options other than :short are not yet implemented in Hologram",
-      );
+    let decimals = null;
+    let scientific = 20; // erlang default
+    let isCompact = false;
+    let isShort = false;
+
+    // Parse options
```
bartblast, line 649: Since you'll use the same logic in `:erlang.float_to_list/2`, we could extract this part to a "virtual" helper function, something like `:erlang._parse_float_opts/1`, and test it in isolation. Just please remember to add this function to the deps in both `erlang.mjs` and `call_graph.ex`.

assets/js/erlang/erlang.mjs
```diff
@@ -589,17 +589,60 @@ const Erlang = {
...
+          Interpreter.isStrictlyEqual(key, Type.atom("scientific")) &&
+          Type.isInteger(value)
+        ) {
+          scientific = Number(value.value);
+        } else {
+          throw new HologramInterpreterError(
+            `:erlang.float_to_binary/2 option {${key.value}, ${value.value}} is not supported in Hologram`,
+          );
+        }
+      } else {
```
bartblast, line 674: This needs to be consistent with Erlang behaviour. Erlang throws and error for not supported options:

```elixir
iex> :erlang.float_to_binary(123.0, [:abc])
** (ArgumentError) errors were found at the given arguments:

  * 2nd argument: invalid option in list

    :erlang.float_to_binary(123.0, [:abc])
    iex:1: (file)
```

assets/js/erlang/erlang.mjs
```diff
@@ -589,17 +589,60 @@ const Erlang = {
...
+    // Parse options
+    for (const opt of opts.data) {
+      if (Interpreter.isStrictlyEqual(opt, Type.atom("short"))) {
+        isShort = true;
+      } else if (Interpreter.isStrictlyEqual(opt, Type.atom("compact"))) {
+        isCompact = true;
+      } else if (Type.isTuple(opt) && opt.data.length === 2) {
+        const key = opt.data[0];
+        const value = opt.data[1];
+        if (
```
bartblast, line 658: We need to be consistent with Erlang behaviour and throw error for invalid decimals and scientific option values, e.g.

```elixir
iex> :erlang.float_to_binary(0.7, [{:scientific, :abc}])
** (ArgumentError) errors were found at the given arguments:

  * 2nd argument: invalid option in list

    :erlang.float_to_binary(0.7, [scientific: :abc])
    iex:2: (file)
```

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1878,10 +1878,54 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
   describe "float_to_binary/2" do
+    test "default option for []" do
+      assert :erlang.float_to_binary(7.12, []) == "7.12000000000000010658e+00"
+    end
+
     test ":short option" do
       assert :erlang.float_to_binary(0.1 + 0.2, [:short]) == "0.30000000000000004"
     end
 
+    test "{:decimals, 0}" do
```
bartblast: Could you make the test names more descriptive by specifying their intent? I'm not sure what specific cases they're testing (beyond testing arbitrary option values). Also, I think it would be valuable to test combinations of options following some pattern, in addition to testing individual options in isolation.

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1878,10 +1878,54 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
   end
 
   describe "float_to_binary/2" do
+    test "default option for []" do
+      assert :erlang.float_to_binary(7.12, []) == "7.12000000000000010658e+00"
+    end
+
     test ":short option" do
       assert :erlang.float_to_binary(0.1 + 0.2, [:short]) == "0.30000000000000004"
```
bartblast: x + y tests the + operator (we don't need it)

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2564,82 +2564,240 @@ describe("Erlang", () => {
   describe("float_to_binary/2", () => {
```
bartblast, line 2742: Comments for the Elixir consistency tests apply here too.

assets/js/erlang/erlang.mjs
bartblast: Got it, thanks.

assets/js/erlang/erlang.mjs
ideaMarcos: are we reusing or is `float_to_list/2` just calling `float_to_binary/2` internally.

ideaMarcos: hey @bartblast if you could comment on above

bartblast: Sorry for the delay @ideaMarcos - I must have missed the notifications!
I initially considered extracting the options parsing logic into a separate helper function, but after looking at this more closely and reviewing your comment, I think it's better to keep things as they are and simply have `:float_to_binary/2` call `:float_to_list/2`. This approach works well for now.

bartblast:
Thank you @ideaMarcos! 💜

PR #394, opened 2025-11-20
Port :lists.min/1 to JS

### LINKED ISSUE #386: Port :lists.min/1 to JS

cheerfulstoic:
Fixes #386

### DIFF
--- a/assets/js/erlang/lists.mjs
+++ b/assets/js/erlang/lists.mjs
@@ -214,6 +214,28 @@ const Erlang_Lists = {
   // End member/2
   // Deps: []
 
+  // Start min/1
+  "min/1": (list) => {
+    if (!Type.isProperList(list) || list.data.length === 0) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":lists.min/1", [list]),
+      );
+    }
+
+    const first = list.data[0];
+    let min = first;
+
+    for (let i = 1; i < list.data.length; i++) {
+      if (Interpreter.compareTerms(list.data[i], min) < 0) {
+        min = list.data[i];
+      }
+    }
+
+    return min;
+  },
+  // End min/1
+  // Deps: []
+
   // Start reverse/1
   "reverse/1": (list) => {
     if (!Type.isList(list)) {
--- a/test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
@@ -373,6 +373,93 @@ defmodule Hologram.ExJsConsistency.Erlang.ListsTest do
     end
   end
 
+  describe "min/1" do
+    test "returns the element from a list of length 1" do
+      assert :lists.min([123]) == 123
+    end
+
+    test "returns the smaller element from a list of size 2 with first being smallest" do
+      assert :lists.min([123, 321]) == 123
+    end
+
+    test "returns the smaller element from a list of size 2 with second being smallest" do
+      assert :lists.min([21, 12]) == 12
+    end
+
+    test "returns the element from a list of size 2 when both are the same" do
+      assert :lists.min([16, 16]) == 16
+    end
+
+    test "comparing same typed values to each other" do
+      assert :lists.min(["16", "2"]) == "16"
+
+      assert :lists.min([:abd, :abc]) == :abc
+    end
+
+    test "comparing different typed values against each other" do
+      all = Enum.take_random(["16", :abc, 16, 16.5], 4)
+
+      assert :lists.min(all) == 16
+
+      without_integer = Enum.take_random(["16", :abc, 16.5], 3)
+
+      assert :lists.min(without_integer) == 16.5
+
+      without_integer_and_float = Enum.take_random(["16", :abc], 2)
+
+      assert :lists.min(without_integer_and_float) == :abc
+    end
+
+    test "returns the smallest element from a large list with many duplicates" do
+      list = [
+        42,
+        82,
+        7,
+        91,
+        23,
+        65,
+        14,
+        23,
+        88,
+        3,
+        56,
+        29,
+        71,
+        18,
+        3,
+        35,
+        82,
+        47
+      ]
+
+      assert :lists.min(list) == 3
+    end
+
+    test "raises FunctionClauseError if the argument is not a list" do
+      expected_msg = build_function_clause_error_msg(":lists.min/1")
+
+      assert_raise FunctionClauseError, expected_msg, fn ->
+        :lists.min(:abc)
+      end
+    end
+
+    test "raises FunctionClauseError if the argument is an improper list" do
+      expected_msg = build_function_clause_error_msg(":lists.min/2")
+
+      assert_raise FunctionClauseError, expected_msg, fn ->
+        :lists.min([1 | 2])
+      end
+    end
+
+    test "raises FunctionClauseError if the argument is an empty list" do
+      expected_msg = build_function_clause_error_msg(":lists.min/1")
+
+      assert_raise FunctionClauseError, expected_msg, fn ->
+        :lists.min([])
+      end
+    end
+  end
+
   describe "reverse/1" do
     test "returns a list with the elements in the argument in reverse order" do
       assert :lists.reverse([1, 2, 3]) == [3, 2, 1]
--- a/test/javascript/erlang/lists_test.mjs
+++ b/test/javascript/erlang/lists_test.mjs
@@ -3,6 +3,7 @@
 import {
   assert,
   assertBoxedError,
+  assertBoxedStrictEqual,
   assertBoxedFalse,
   assertBoxedTrue,
   contextFixture,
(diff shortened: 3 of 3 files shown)


cheerfulstoic:
Realized I forgot to run `mix f`, but when I did I got this:

```
** (Mix) Unknown dependency :hologram given to :import_deps in the formatter configuration. Make sure the dependency is listed in your mix.exs for environment :dev and you have run "mix deps.get"
** (exit) 1
    (mix 1.18.2) lib/mix/tasks/cmd.ex:74: Mix.Tasks.Cmd.run/1
    (mix 1.18.2) lib/mix/task.ex:495: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.18.2) lib/mix/task.ex:561: Mix.Task.run_alias/6
    (mix 1.18.2) lib/mix/cli.ex:107: Mix.CLI.run_task/2
    /Users/brian/.asdf/installs/elixir/1.18.2-otp-27/bin/mix:2: (file)
```

Just signing off for the day now, but if you don't know what that's about I can look later.  `mix check` passes, though

brGuirra:
@cheerfulstoic I was having the same issue. You need to go to `test/features` and run `mix deps.get` there to install the missing dependencies. It's a like another project inside the main repo with its separete configuration and dependencies.

bartblast:
@cheerfulstoic,
@brGuirra is correct. I also noticed that `npm install` is required, so I created a simple mix task that automates the entire setup process. See: https://github.com/bartblast/hologram/commit/4b1e39b42230c9ad0b51cd3976a04354b4fe8bbb
You can now run:
`$ mix setup`
I've also added an "Install Dependencies" section to the [Contributing Guide](https://hologram.page/reference/contributing).

cheerfulstoic:
Thank you! I synced, rebased, and ran `mix setup`, `mix check`, and `mix f`.  I've committed the changes from `mix f`, so hopefully it's good to go!

assets/js/erlang/lists.mjs
```diff
@@ -214,6 +214,31 @@ const Erlang_Lists = {
   // End member/2
   // Deps: []
 
+  // Start min/1
+  "min/1": (list) => {
+    if (!Type.isProperList(list)) {
```
bartblast: Since the errors are the same for both of these cases you could do:
`if (!Type.isProperList(list) || list.data.length === 0)`

assets/js/erlang/lists.mjs
```diff
@@ -214,6 +214,31 @@ const Erlang_Lists = {
...
+    if (list.data.length === 0) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":lists.min/1", [list]),
+      );
+    }
+
+    const first = list.data[0];
+    const rest = list.data.slice(1);
+
+    return rest.reduce(
```
bartblast: This approach:

```
const first = list.data[0];
let min = first;

for (let i = 1; i < list.data.length; i++) {
  if (Interpreter.compareTerms(list.data[i], min) < 0) {
    min = list.data[i];
  }
}

return min;
```

* avoids `slice(1)` which creates a new array copying all elements except the first
* eliminates the function call overhead of reduce
* more cache-friendly with direct array indexing
* faster especially for larger arrays

test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
```diff
@@ -373,6 +373,69 @@ defmodule Hologram.ExJsConsistency.Erlang.ListsTest do
...
+
+    test "returns the smaller element from a list of size 2 with second being smallest" do
+      assert :lists.min([21, 12]) == 12
+    end
+
+    test "returns the element from a list of size 2 when both are the same" do
+      assert :lists.min([16, 16]) == 16
+    end
+
+    test "comparing other types to each other" do
```
bartblast: This one is redundant IMO.

test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
```diff
@@ -373,6 +373,69 @@ defmodule Hologram.ExJsConsistency.Erlang.ListsTest do
...
+      assert :lists.min(["16", "2"]) == "16"
+
+      assert :lists.min([:abd, :abc]) == :abc
+    end
+
+    test "comparing different types" do
```
bartblast: Instead, could you create a list containing items of the following types: atom, bitstring, float, integer, pid, and tuple? Then shuffle the items randomly (so each test run has a different order), apply them as input to the tested function, and verify that structural comparison is applied correctly?

test/javascript/erlang/lists_test.mjs
```diff
@@ -669,42 +670,173 @@ describe("Erlang_Lists", () => {
     });
   });
 
-  describe("reverse/1", () => {
-    const reverse = Erlang_Lists["reverse/1"];
+  (describe("min/1", () => {
```
bartblast: It looks like the JS tests got messed up.

This is incorrect: `(describe("min/1", () => { ` - there should be no `(` before `describe`, and this caused this strange diff.
 
Please see how other describe blocks are structured.

test/javascript/erlang/lists_test.mjs
```diff
@@ -669,42 +670,173 @@ describe("Erlang_Lists", () => {
...
+      );
+
+      assertBoxedError(
+        () => min(Type.list([])),
+        "FunctionClauseError",
+        expectedMessage,
+      );
+    });
+  }),
+    describe("reverse/1", () => {
```
bartblast: Same here, this is strangely indented and probably is another cause of the strange diff.

test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
cheerfulstoic: I'm wondering... if I put ["16", 16, :abc, 16.5] into a list, I'll always get `16` as the result of `:lists.min/1`, but that doesn't tell me anything about which is the `min` between "16" and `:abc` or `:abc` and 16.5, right?

I mean, maybe there don't need to be so many tests because we know that the function is using `Interpreter.compareTerms` and we just need to do a few sanity checks? 🤷

cheerfulstoic: Also, for shuffling, I don't think there's a simple way to do that in JS, is there? Unless maybe you have a helper?

test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
cheerfulstoic: I'm happy to remove it, but just checking: redundant to what?  Because we're comparing integers in other tests?

cheerfulstoic:
Thanks! Some fixes, some questions.  Hopefully I can wrap it up in the next round and then knock out `:lists.max/1` 👍

test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
bartblast: Sorry, the test name misled me a little bit. I thought that this test checks the same thing as the “comparing different types” test does. Could you change it to indicate that it compares the same types?

test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs
bartblast: > I'm wondering... if I put ["16", 16, :abc, 16.5] into a list, I'll always get 16 as the result of :lists.min/1, but that 
doesn't tell me anything about which is the min between "16" and :abc or :abc and 16.5, right?

Yeah, that's why you need to shuffle it. What I'm suggesting will test that the structural comparison (through the `compareTerms()` function) is applied correctly, so we don't need to use so many cases.

> Also, for shuffling, I don't think there's a simple way to do that in JS, is there? Unless maybe you have a helper?

Something like this will do:
`const shuffled = array.sort(() => Math.random() - 0.5)`

bartblast:
@cheerfulstoic I replied to your questions, but I also just noticed that the error case tests are missing in the Elixir consistency tests. The goal is to have identical JS and Elixir tests so that we: 1) verify the client/server behavior is the same, and 2) are alerted whenever something changes with OTP updates.

cheerfulstoic:
Ok, changes made, thank you!

> I also just noticed that the error case tests are missing in the Elixir consistency tests.

Are you sure? 🤔 😅  Do you mean in `test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs`? I have them in alphabetical order too, I think (just after `member/2`)...

bartblast:
> > I also just noticed that the error case tests are missing in the Elixir consistency tests.
> 
> Are you sure? 🤔 😅 Do you mean in `test/elixir/hologram/ex_js_consistency/erlang/lists_test.exs`? I have them in alphabetical order too, I think (just after `member/2`)...

You're missing error case tests in Elixir. For example, you have a "raises FunctionClauseError if the argument is an empty list" test in JS, but no corresponding test in Elixir.

cheerfulstoic:
Ah, right 🤦  Easy enough, fixed!

bartblast:
Thank you, @cheerfulstoic! 🙂

PR #401, opened 2025-11-21
Port :erlang.binary_to_list/1 to JS

jerry-chao:
Fixes [#388](https://github.com/bartblast/hologram/issues/388)

### DIFF
--- a/assets/js/erlang/erlang.mjs
+++ b/assets/js/erlang/erlang.mjs
@@ -426,13 +426,7 @@ const Erlang = {
 
     Bitstring.maybeSetBytesFromText(binary);
     const bytes = binary.bytes;
-    const result = [];
-
-    for (let i = 0; i < bytes.length; i++) {
-      result.push(Type.integer(BigInt(bytes[i])));
-    }
-
-    return Type.list(result);
+    return Type.list(Array.from(bytes).map((b) => Type.integer(b)));
   },
   // End binary_to_list/1
   // Deps: []
--- a/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
@@ -1479,10 +1479,6 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
       assert :erlang.binary_to_list(<<>>) == []
     end
 
-    test "converts a binary with mixed byte values" do
-      assert :erlang.binary_to_list(<<0, 127, 255>>) == [0, 127, 255]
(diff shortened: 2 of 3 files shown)


assets/js/erlang/erlang.mjs
```diff
@@ -442,6 +416,27 @@ const Erlang = {
...
+
+    Bitstring.maybeSetBytesFromText(binary);
+    const bytes = binary.bytes;
+    const result = [];
+
+    for (let i = 0; i < bytes.length; i++) {
+      result.push(Type.integer(BigInt(bytes[i])));
+    }
+
+    return Type.list(result);
```
bartblast: This will be both simpler and more performant:
`return Type.list(bytes.map(b => Type.integer(b)));`

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1638,6 +1466,36 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+
+    test "converts a text-based binary to a list of integers" do
+      assert :erlang.binary_to_list("ABC") == [65, 66, 67]
+    end
+
+    test "converts an empty binary to an empty list" do
+      assert :erlang.binary_to_list(<<>>) == []
+    end
+
+    test "converts a binary with mixed byte values" do
```
bartblast: This may be redundant - can you explain its goal?

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2174,6 +1824,70 @@ describe("Erlang", () => {
...
+
+    it("converts an empty binary to an empty list", () => {
+      const binary = Type.bitstring("");
+      const result = binary_to_list(binary);
+      const expected = Type.list([]);
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it("converts a binary with mixed byte values", () => {
```
bartblast: This may be redundant - can you explain its goal?

assets/js/erlang/erlang.mjs
jerry-chao: it is perfect, i will change this

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
jerry-chao: this is for the binary which is not standard utf-8 encode binary , maybe it is not necessary for this function

bartblast:
Thank you for help, @jerry-chao! :)

PR #405, opened 2025-11-21
Port :erlang.trunc/1 to JS

tenkiller:
Closes https://github.com/bartblast/hologram/issues/404


assets/js/erlang/erlang.mjs
```diff
@@ -1029,6 +1013,28 @@ const Erlang = {
...
+    let value = number.value;
+
+    if (Type.isFloat(number)) {
+      value = Math.trunc(value);
+    }
+
+    // Erlang trunc/1 converts -0 to 0 while native JavasScript does not
+    value = Object.is(value, -0) ? 0 : value;
+
+    return Type.integer(value);
```
bartblast: This should be simpler and faster:

```
if (Type.isFloat(number)) {
  return Type.integer(Math.trunc(number.value) + 0)
}

return number;
```

Please keep your original comment about signed 0.

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -2358,6 +2336,50 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+
+    test "keeps negative integer unchanged" do
+      assert :erlang.trunc(-1) == -1
+    end
+
+    test "keeps zero integer unchanged" do
+      assert :erlang.trunc(0) == 0
+    end
+
+    test "converts negative zero integer to positive zero integer" do
```
bartblast, line 2564: Not strictly necessary (knowing the internals) but worth adding signed plus 0 integer test for symmetry.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -3436,6 +3404,74 @@ describe("Erlang", () => {
...
+      assert.deepStrictEqual(result, integer);
+    });
+
+    it("keeps zero integer unchanged", () => {
+      const result = testedFun(integer0);
+
+      assert.deepStrictEqual(result, integer0);
+    });
+
+    it("converts negative zero integer to positive zero integer", () => {
```
bartblast, line 3739: Not strictly necessary (knowing the internals) but worth adding signed plus 0 integer test for symmetry.

bartblast:
Thank you @tenkiller! :)

PR #411, opened 2025-11-23
Port :erlang.float/1 to JS

washbin:
Closes the issue #410

### DIFF
--- a/assets/js/erlang/erlang.mjs
+++ b/assets/js/erlang/erlang.mjs
@@ -571,13 +571,15 @@ const Erlang = {
 
   // Start float/1
   "float/1": (number) => {
-    if (!Type.isNumber(number)) {
-      Interpreter.raiseArgumentError(
-        Interpreter.buildArgumentErrorMsg(1, "not a number"),
-      );
+    if (Type.isInteger(number)) {
+      return Type.float(Number(number.value));
+    } else if (Type.isFloat(number)) {
+      return number;
     }
 
-    return Type.float(Number(number.value));
+    Interpreter.raiseArgumentError(
+      Interpreter.buildArgumentErrorMsg(1, "not a number"),
+    );
   },
   // End float/1
   // Deps: []
--- a/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
@@ -1877,19 +1877,19 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
-  describe "float/2" do
+  describe "float/1" do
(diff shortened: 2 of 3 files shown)


test/javascript/erlang/erlang_test.mjs
```diff
@@ -1,5 +1,10 @@
 "use strict";
 
+import Bitstring from "../../../assets/js/bitstring.mjs";
```
bartblast: Please keep the same structure of imports as in other test files.

assets/js/erlang/erlang.mjs
```diff
@@ -569,6 +569,19 @@ const Erlang = {
   // End error/2
   // Deps: []
 
+  // Start float/1
+  "float/1": (number) => {
+    if (!Type.isNumber(number)) {
```
bartblast: This should be faster

```js
if (Type.isInteger(number)) {
  return Type.float(Number(number.value));
} else if (Type.isFloat(number)) {
  return number;
}

  Interpreter.buildArgumentErrorMsg(1, "not a number"),
);
```

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1877,6 +1877,22 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "float/2" do
```
bartblast: Could you add a test for integer input?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1877,6 +1877,22 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+    test "returns float for integer" do
+      assert :erlang.float(123) == 123.0
+    end
+
+    test "idempotent for float" do
```
bartblast: This is redundant IMO (it tests the `+` operator) in practice.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2561,6 +2560,36 @@ describe("Erlang", () => {
     assertBoxedError(() => error(reason, args), "MyError", "my message");
   });
 
+  describe("float/1", () => {
```
bartblast, line 2564: Comments for the Elixir tests apply here too.

bartblast (changes requested):

washbin:
Can you check now @bartblast , I have fixed most of what you suggested, as for the integer input test part I am a little confused, do you mean to add more tests for integers ?
The formatting was due to a editor misconfig it seems.

bartblast:
Looks good, thanks for help @washbin! :)

Issue #296, opened 2025-11-16
Port :sets.from_list/2 to JS


bartblast:
Part of the Elixir → JavaScript porting initiative (https://hologram.page/blog/elixir-to-javascript-porting-initiative).

@Lucassifoni is working on this.

Lucassifoni:
Commenting to assign

Lucassifoni:
I need to implement :maps.from_keys/2 to be able to have consistent errors between elixir and js tests. Opening an issue and sending a PR for :maps.from_keys.

PR #416, opened 2025-11-25
Port :sets.from_list/2 and :sets.new/1 to JS

Lucassifoni:
A few remarks : 

- In test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs, I define a SetsTestHelper module normalizing version discrepancies. It is only used by functions that take in the version argument. Is that alright for you ?
- I did not have a great idea yet on how to mimic the rejection of version 1 sets elixir-side in the consistency tests, and left a warning in test/javascript/erlang/sets_test.mjs

Feel free to ask for any relevant changes and I will proceed working on them.

see #296 

Best,
Lucas

## Summary by CodeRabbit

* **New Features**
  * Added set creation from lists and an explicit new(empty) set constructor with option validation.
  * Exposed clearer versioned options handling (supports version 2; rejects unsupported versions).

* **Bug Fixes**
  * Improved option validation and more precise error reporting for invalid or non-list options.

* **Tests**
  * Expanded JavaScript and Elixir tests covering construction, duplicates, empty sets, and error cases.

### DIFF
--- a/assets/js/erlang/sets.mjs
+++ b/assets/js/erlang/sets.mjs
@@ -1,5 +1,6 @@
 "use strict";
 
+import Erlang_Lists from "./lists.mjs";
 import Erlang_Maps from "./maps.mjs";
 import HologramInterpreterError from "../errors/interpreter_error.mjs";
 import Interpreter from "../interpreter.mjs";
@@ -12,44 +13,32 @@ import Type from "../type.mjs";
 const Erlang_Sets = {
   // Start _validate_opts/3
   "_validate_opts/3": (opts, source, args) => {
-    const raiseFunctionClauseError = () => {
+    if (!Type.isList(opts)) {
       Interpreter.raiseFunctionClauseError(
         Interpreter.buildFunctionClauseErrorMsg(source, args),
       );
-    };
-
-    if (!Type.isList(opts)) {
-      raiseFunctionClauseError();
     }
 
-    if (opts.data.length === 0) {
+    const versionOpt = Erlang_Lists["keyfind/3"](
+      Type.atom("version"),
+      Type.integer(1),
+      opts,
+    );
+
+    if (Type.isBoolean(versionOpt) && versionOpt.value === "false") {
       return;
     }
 
-    if (opts.data.length !== 1) {
-      raiseFunctionClauseError();
-    }
-
-    const opt = opts.data[0];
-
-    if (!Type.isTuple(opt) || opt.data.length !== 2) {
-      raiseFunctionClauseError();
-    }
-
-    const [key, value] = opt.data;
-
-    if (!Type.isAtom(key) || key.value !== "version") {
-      raiseFunctionClauseError();
-    }
+    const value = versionOpt.data[1];
 
     if (!Type.isInteger(value)) {
-      raiseFunctionClauseError();
+      return;
     }
 
     const version = Number(value.value);
 
-    if (version !== 1 && version !== 2) {
-      raiseFunctionClauseError();
+    if (version === 2) {
+      return;
     }
 
     if (version === 1) {
@@ -57,9 +46,11 @@ const Erlang_Sets = {
         ":sets version 1 is not supported in Hologram, use [{:version, 2}] option",
       );
     }
+
+    Interpreter.raiseCaseClauseError(value);
   },
   // End _validate_opts/3
-  // Deps: []
+  // Deps: [:lists.keyfind/3]
 
   // Start from_list/2
   "from_list/2": (list, opts) => {
--- a/test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
@@ -43,6 +43,16 @@ defmodule Hologram.ExJsConsistency.Erlang.SetsTest do
       result = :sets.new([{:version, 2}])
       assert result == %{}
     end
+
+    test "ignores invalid option keys (doesn't raise)" do
+      :sets.new([{:invalid, 2}])
+    end
+
+    test "raises CaseClauseError when version is invalid" do
+      assert_raise CaseClauseError, fn ->
+        :sets.new([{:version, 3}])
+      end
+    end
   end
 
   describe "to_list/1" do
--- a/test/javascript/erlang/sets_test.mjs
+++ b/test/javascript/erlang/sets_test.mjs
@@ -91,6 +91,23 @@ describe("Erlang_Sets", () => {
       assert.deepStrictEqual(result, expected);
     });
 
+    it("creates a new empty set with empty opts (defaults to version 2)", () => {
+      const result = new_1(Type.list([]));
+      const expected = Type.map([]);
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it("ignores invalid option keys", () => {
+      const opts = Type.list([
+        Type.tuple([Type.atom("invalid"), Type.integer(2)]),
+      ]);
+      const result = new_1(opts);
+      const expected = Type.map([]);
+
+      assert.deepStrictEqual(result, expected);
+    });
+
     it("raises FunctionClauseError when opts is not a list", () => {
       const opts = Type.atom("invalid");
 
@@ -101,7 +118,7 @@ describe("Erlang_Sets", () => {
       );
     });
 
-    it("raises HologramInterpreterError when version is not 2", () => {
+    it("raises HologramInterpreterError when version is 1", () => {
       const opts = Type.list([
         Type.tuple([Type.atom("version"), Type.integer(1)]),
       ]);
@@ -110,6 +127,17 @@ describe("Erlang_Sets", () => {
         () => new_1(opts),
         HologramInterpreterError,
         ":sets version 1 is not supported in Hologram, use [{:version, 2}] option",
+      );
+    });
+
+    it("raises CaseClauseError when version is invalid", () => {
+      const version = Type.integer(3);
+      const opts = Type.list([Type.tuple([Type.atom("version"), version])]);
+
+      assertBoxedError(
+        () => new_1(opts),
+        "CaseClauseError",
+        "no case clause matching: " + Interpreter.inspect(version),
       );
     });
   });
--- a/lib/hologram/compiler/call_graph.ex
+++ b/lib/hologram/compiler/call_graph.ex
@@ -39,6 +39,7 @@ defmodule Hologram.Compiler.CallGraph do
     {{:maps, :get, 2}, {:maps, :get, 3}},
     {{:maps, :update, 3}, {:maps, :is_key, 2}},
     {{:maps, :update, 3}, {:maps, :put, 3}},
+    {{:sets, :_validate_opts, 3}, {:lists, :keyfind, 3}},
     {{:sets, :from_list, 2}, {:maps, :from_keys, 2}},
     {{:sets, :from_list, 2}, {:sets, :_validate_opts, 3}},
     {{:sets, :new, 1}, {:sets, :_validate_opts, 3}},


Lucassifoni:
I'm starting to think (well it crossed my mind a few times this last 10 days) that I should bite the bullet and implement sets v1, so we avoid introducing exceptions, special cases and helpers. 
What do you think ?

assets/js/erlang/sets.mjs
```diff
@@ -0,0 +1,73 @@
...
+  // Start from_list/2
+  "from_list/2": (list, opts) => {
+    validateErlangSetsOpts(opts, ":sets.from_list/2", [list, opts]);
+    return Erlang_Maps["from_keys/2"](list, Type.list([]));
+  },
+  // End from_list/2
+  // Deps: [:maps.from_keys/2]
+};
+
+export default Erlang_Sets;
```
bartblast: That's a good idea, but we need to treat it as a "virtual" Erlang function, e.g., `:sets._validate_opts/3`. Here's why: Hologram extracts code between `Start/End` delimiters for ported modules and only sees what it's explicitly told to include - which is why we defined the dependencies in comments within the call graph. The way you've used it (as a const outside the function) means Hologram wouldn't be able to detect it. Another option would be to put helper functions into JS runtime classes (similar to `Type`, `Interpreter`, etc.), but that would mean including these helpers in all Hologram applications, even when they're not actually used.

So, you'll need to implement it as a "virtual" function and add it to Deps in any functions that reference it (remember about `@erlang_mfa_edges` in `call_graph.ex` too).

assets/js/erlang/sets.mjs
```diff
@@ -0,0 +1,73 @@
...
+
+const validateErlangSetsOpts = (opts, source, args) => {
+  if (!Type.isList(opts)) {
+    Interpreter.raiseFunctionClauseError(
+      Interpreter.buildFunctionClauseErrorMsg(source, args),
+    );
+  }
+
+  let version = 2;
+  for (const opt of opts.data) {
```
bartblast: Looking at the Erlang docs (https://www.erlang.org/doc/apps/stdlib/sets.html#new/1 and https://www.erlang.org/doc/apps/stdlib/sets.html#from_list/2), we need to validate that `opts` is a single-element tuple containing a `:version` key with a value of either `1` or `2`. I don't think a for loop is necessary here.
Note that we should raise an error (matching the server-side behavior) when the version value is invalid or when `opts` contains keys other than `:version`. This ensures consistency between client and server implementations, and failing tests will alert us if the server behavior changes.

assets/js/erlang/sets.mjs
```diff
@@ -0,0 +1,73 @@
...
+
+import Erlang_Maps from "./maps.mjs";
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const validateErlangSetsOpts = (opts, source, args) => {
```
bartblast: That's a good idea, but we need to treat it as a "virtual" Erlang function, e.g., `:sets._validate_opts/3`. Here's why: Hologram extracts code between `Start/End` delimiters for ported modules and only sees what it's explicitly told to include - which is why we defined the dependencies in comments and within the call graph. The way you've used it (as a const outside the function) means Hologram wouldn't be able to detect it. Another option would be to put helper functions into JS runtime classes (similar to `Type`, `Interpreter`, etc.), but that would mean including these helpers in all Hologram applications, even when they're not actually used.

So, you'll need to implement it as a "virtual" function and add it to Deps in any functions that reference it (remember about `@erlang_mfa_edges` in `call_graph.ex` too).

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -0,0 +1,120 @@
...
+      assert result == %{1 => [], 2 => [], 3 => []}
+    end
+
+    test "creates a set from a list with duplicate elements" do
+      result = SetsTestHelper.from_list([1, 2, 1, 3])
+      assert result == %{1 => [], 2 => [], 3 => []}
+    end
+  end
+
+  describe "from_list/2" do
```
bartblast: We also need to test `opts` validation (version other than 1 or 2, keys other than `:version`).

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -0,0 +1,120 @@
...
+    def from_list(list, options), do: :sets.from_list(list, options)
+  end
+else
+  if otp_version < 24 do
+    # We do not support OTP < 24, because Hologram needs elixir >= 1.15 for the ~HOLO sigil
+    # and multi-letter sigils were introduced as of elixir 1.15. There is no OTP version under
+    # 24 that supports elixir 1.15 so we can safely disregard it here.
+  else
+    # OTP 24 to 27, we normalize calls to new/0 and from_list/{1,2}
+    defmodule SetsTestHelper do
```
bartblast: I reviewed the client runtime reference and the MapSet module, and it appears that neither `new/0` nor `from_list/1` are necessary. Since Elixir 1.15, the [version: 2] option has been used consistently in Elixir modules. I believe we can safely remove these functions to resolve the normalization issue. However, the version == 2 validation should remain in place.

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -0,0 +1,120 @@
...
+    test "creates a set from an empty list with version 2 option" do
+      result = SetsTestHelper.from_list([], [{:version, 2}])
+      assert result == %{}
+    end
+
+    test "creates a set from a list with elements and version 2 option" do
+      result = SetsTestHelper.from_list([1, 2, 3], [{:version, 2}])
+      assert result == %{1 => [], 2 => [], 3 => []}
+    end
+
```
bartblast, line 27: Could you add a test case for duplicate items?

test/javascript/erlang/sets_test.mjs
```diff
@@ -0,0 +1,163 @@
...
+  describe("new/0", () => {
+    it("creates a new empty set", () => {
+      const result = Erlang_Sets["new/0"]();
+      const expected = Type.map([]);
+
+      assert.deepStrictEqual(result, expected);
+    });
+  });
+
+  describe("new/1", () => {
```
bartblast, line 84: Could you assign the tested function to a const at the describe block level (like in other tests)?

this can be:
`const new_1 = Erlang_Sets["new/1"]
`or
`const testedFun = Erlang_Sets["new/1"]`

(this applies also to other functions)

test/javascript/erlang/sets_test.mjs
```diff
@@ -0,0 +1,163 @@
...
+    it("creates a new empty set with version 2 option", () => {
+      const opts = Type.list([
```
bartblast: Since `opts` will always be `[version: 2]` it makes sense to specify a const a the beginning of the file.

test/javascript/erlang/sets_test.mjs
```diff
@@ -0,0 +1,163 @@
...
+      );
+    });
+
+    it("raises FunctionClauseError when opts is not a list", () => {
+      const opts = Type.atom("invalid");
+
+      assertBoxedError(
+        () => Erlang_Sets["new/1"](opts),
+        "FunctionClauseError",
+        "no function clause matching in :sets.new/1\n\nThe following arguments were given to :sets.new/1:\n\n    # 1\n    :invalid\n",
```
bartblast: Could you use `buildFunctionClauseErrorMsg()` helper?

test/javascript/erlang/sets_test.mjs
```diff
@@ -0,0 +1,163 @@
...
+      );
+    });
+    // End Hologram-specific limitation tests (no Elixir counterpart):
```
bartblast: Shouldn't this comment go after this test: "raises ArgumentError when version is not 2"?

.tool-versions
```diff
@@ -1,3 +1,3 @@
 elixir 1.18.2-otp-27
 erlang 27.2.4
-nodejs 22.9.0
```
bartblast: Looks like this slipped through accidentally.

Could you:
`$ git checkout .tool-versions`

assets/js/erlang/sets.mjs
```diff
@@ -0,0 +1,73 @@
...
+      if (Type.isAtom(key) && key.value === "version") {
+        if (Type.isInteger(value)) {
+          version = Number(value.value);
+        }
+      }
+    }
+  }
+
+  if (version !== 2) {
+    Interpreter.raiseArgumentError(
```
bartblast: Could you raise `HologramInterpreterError` (this is specific to client-side Hologram runtime only).

bartblast:
> I'm starting to think (well it crossed my mind a few times this last 10 days) that I should bite the bullet and implement sets v1, so we avoid introducing exceptions, special cases and helpers. What do you think ?

Take a look at my comment here: https://github.com/bartblast/hologram/pull/416#discussion_r2582756934

Lucassifoni:
Thank you for this thorough review, I’ll be able to make the changes quickly.Le 2 déc. 2025 à 22:15, Bart Blast ***@***.***> a écrit :﻿bartblast left a comment (bartblast/hologram#416)

I'm starting to think (well it crossed my mind a few times this last 10 days) that I should bite the bullet and implement sets v1, so we avoid introducing exceptions, special cases and helpers. What do you think ?

Take a look at my comment here: #416 (comment)

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
Lucassifoni: Hi Bart,
I needed this to be able to assert on data shapes being maps JS-side as well as Elixir-side under OTP 28, when calling :sets directly, not through Elixir (which specifically adds the version option). If you think we can go up a level of abstraction in tests I could effectively remove it.

bartblast: Hi Lucas, could you elaborate a bit on what you mean? I’m not fully following your point yet and want to make sure I understand it correctly. Thanks!

Lucassifoni: Hi Bart,
The issue with :sets from my point of view is that we are going to implement it as a map data shape only in Hologram's JS implementation, but even though the Elixir standard library that leverages :sets passes the {:version, 2} option, if an user has business code that directly uses the :sets erlang module, it will default to version 1 from OTP 24 to 27. OTP 28 defaults to version 2.

So if I write tests for :sets, I will get back a tuple bucket data structure from OTP 24 to 27, or a map starting from OTP 28, if I do not force the version 2 option. 

My thinking is then : if we want to assert on data shapes, we need to have sets tests going through some kind of normalization..

Maybe I'm missing something, let me know if this reasoning does not make sense.

bartblast: Lucas, you're right - if user Elixir code directly calls the Erlang `:sets` module without specifying the version option, it would default to version 1 on OTP 24-27. However, my point is that we shouldn't port `:sets.new/0` and `:sets.from_list/1` (which use default version values) since they're not needed by the Elixir stdlib (looking at the client runtime reference). Hologram is designed to support Elixir code - we don't advertise full Erlang support (not yet, at least). So I think we need to choose our battles here and focus on what Elixir's standard library actually uses rather than covering all possible Erlang scenarios.

Instead, we should raise a clear error in `:sets.new/1` and `:sets.from_list/2` when version 1 is explicitly passed or when an empty opts list is passed (which would default to version 1 as well). As you noted, this could only happen if some code deliberately uses the `:sets` module directly (not through `MapSet`, which specifically uses version 2). While it's possible, I think it's highly unlikely - and even if it does occur, we can raise a clear error message to notify the client code that Hologram only supports version 2.

This way we only need to test cases that use version 2 maps underneath for both JS tests and Elixir consistency tests. It probably makes sense to document that as well.

I really appreciate your thoroughness and careful consideration of different scenarios though :)

What do you think?

Lucassifoni:
Edit : sorry, reply by email messes too much with the formatting..

Hi Bart,I understand your position on this, it will make things simpler. I will document it and make sure we emit clear error messages. 

So would it make sense to you to write the tests of the :sets module in terms of MapSet which will perform this automatic funneling to version 2 ?

This way we only test :sets through elixir’s stdlib and stay within the bounds of a future automatic correlation of Elixir/JS consistency test results.

Thanks a lot for your time on this matter !

Envoyé de mon iPhoneLe 9 déc. 2025 à 23:23, Bart Blast ***@***.***> a écrit :﻿
@bartblast commented on this pull request.



In test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs:
> +    def new(options), do: :sets.new(options)
+
+    @spec from_list(list()) :: :sets.set(term())
+    def from_list(list), do: :sets.from_list(list)
+
+    @spec from_list(list(), keyword()) :: :sets.set(term())
+    def from_list(list, options), do: :sets.from_list(list, options)
+  end
+else
+  if otp_version < 24 do
+    # We do not support OTP < 24, because Hologram needs elixir >= 1.15 for the ~HOLO sigil
+    # and multi-letter sigils were introduced as of elixir 1.15. There is no OTP version under
+    # 24 that supports elixir 1.15 so we can safely disregard it here.
+  else
+    # OTP 24 to 27, we normalize calls to new/0 and from_list/{1,2}
+    defmodule SetsTestHelper do

Lucas, you're right - if user Elixir code directly calls the Erlang :sets module without specifying the version option, it would default to version 1 on OTP 24-27. However, my point is that we shouldn't port :sets.new/0 and :sets.from_list/1 (which use default version values) since they're not needed by the Elixir stdlib (looking at the client runtime reference). Hologram is designed to support Elixir code - we don't advertise full Erlang support (not yet, at least). So I think we need to choose our battles here and focus on what Elixir's standard library actually uses rather than covering all possible Erlang scenarios.
Instead, we should raise a clear error in :sets.new/1 and :sets.from_list/2 when version 1 is explicitly passed or when an empty opts list is passed (which would default to version 1 as well). As you noted, this could only happen if some code deliberately uses the :sets module directly (not through MapSet, which specifically uses version 2). While it's possible, I think it's highly unlikely - and even if it does occur, we can raise a clear error message to notify the client code that Hologram only supports version 2.
This way we only need to test cases that use version 2 maps underneath for both JS tests and Elixir consistency tests. It probably makes sense to document that as well.
I really appreciate your thoroughness and careful consideration of different scenarios though :)
What do you think?

bartblast:
Hey Lucas!

Just to clarify - we need to keep testing the Erlang functions directly (like `:sets.from_list/2`) in the consistency tests, not `MapSet`. The conventions are important to maintain here because we need to catch any inconsistencies between the implementations now and in the future if something changes in OTP.

Also, with the approach I described, all functions should output maps across all OTP versions - so we're consistently dealing with version 2 format everywhere.

For the version 1 exception test, could you add a `describe("client-only behaviour")` block in the JS test?

If any of this is unclear, just let me know and I'm happy to write it myself.

Lucassifoni:
> Hey Lucas!
> 
> Just to clarify - we need to keep testing the Erlang functions directly (like `:sets.from_list/2`) in the consistency tests, not `MapSet`. The conventions are important to maintain here because we need to catch any inconsistencies between the implementations now and in the future if something changes in OTP.
> 
> Also, with the approach I described, all functions should output maps across all OTP versions - so we're consistently dealing with version 2 format everywhere.
> 
> For the version 1 exception test, could you add a `describe("client-only behaviour")` block in the JS test?
> 
> If any of this is unclear, just let me know and I'm happy to write it myself.

Hello Bart, I think we're on the same page now. I'm wrapping this up.

Lucassifoni:
I think this is simpler and better now that all the :sets version dance is removed.

bartblast:
@coderabbitai full review

bartblast:
Hi Lucas,

Could you pull the new changes from the `dev` branch and address the issues detected by CodeRabbit?

Regarding the CodeRabbit comments, it correctly detected the edge cases that need handling, but the expected behavior should be different than what it suggested:

1. **Empty opts** (`[]`) - should **default to version 2** (don't raise an error)
2. **Invalid version** (e.g., `[{:version, 3}]`) - should raise **CaseClauseError** (not FunctionClauseError)
3. **Invalid option keys** (e.g., `[{:invalid, 2}]`) - should **ignore them** and not raise at all

Lucassifoni:
@coderabbitai this change cannot be included elixir-side, since the output of sets.new/1 without an explicit {:version, 2} tuple is not stable. :
```
Add test for empty opts list defaulting to version 2.

Per bartblast's comment, empty opts [] should default to version 2 (not raise). The JS test has this case covered (line 94-99), but the Elixir consistency test is missing it.
🔎 Proposed test case

   describe "new/1" do
     test "creates a new empty set with version 2 option" do
       result = :sets.new([{:version, 2}])
       assert result == %{}
     end
 
+    test "creates a new empty set with empty opts (defaults to version 2)" do
+      result = :sets.new([])
+      assert result == %{}
+    end
+
     test "ignores invalid option keys (doesn't raise)" do
       :sets.new([{:invalid, 2}])
     end
```

bartblast:
Thanks for your help Lucas! :) 🎉

Lucassifoni:
Now that the annoying part (versions) is done I'll continue implementing :sets. Thanks for your patience @bartblast, I have a better picture of hologram and should be quicker now.

bartblast:
Sounds good! :)

PR #423, opened 2025-11-27
Port :string.titlecase/1 to JS

### LINKED ISSUE #422: Port :string.titlecase/1 to JS
https://www.erlang.org/doc/apps/stdlib/string.html#titlecase/1

steefmin:
## What

- ports :string.titlecase/1 to JS
- closes #422 

## To do

- [x] add test cases for:
  - [x] invalid input types
  - [x] non-charlist lists (handled by blacklist)
  - [x] non-latin chars (handled by blacklist)

## Summary by CodeRabbit

* **New Features**
  * Added titlecase string conversion for text and charlist/list inputs, handling Unicode (including ß -> Ss) and emoji sequences; returns appropriate errors for unsupported inputs.

* **Tests**
  * Added comprehensive JS and Elixir tests covering empty input, simple and idempotent capitalization, Unicode and emoji edge cases, nested lists/charlists, and input immutability.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/string_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/string_test.exs
@@ -33,11 +33,12 @@ defmodule Hologram.ExJsConsistency.Erlang.StringTest do
     end
 
     test "emoji" do
-      [
+      data = [
         {"👩‍👩‍👦‍👦", "👩‍👩‍👦‍👦"},
         {"👩‍🚒", "👩‍🚒"}
       ]
-      |> Enum.each(fn {input, expected} ->
+
+      Enum.each(data, fn {input, expected} ->
         assert :string.titlecase(input) == expected
       end)
     end


bartblast:
Hey @steefmin, thanks for the PR! Is it ready for review?

steefmin:
This first try implementation might work for 90% of the cases, but it seems that Erlang's titlecase has a bunch of edge cases that are not handled yet. Feel free to review what I committed already as it might point me in a better direction, but I understand if you don't want to spend time on a partial solution.

bartblast:
Could you elaborate on the edge cases you’re referring to? More detail would help guide next steps.

bartblast:
@steefmin Could you elaborate on the edge cases you’re referring to? More detail would help guide next steps.

steefmin:
@bartblast you can find some in the tests, but there are a bunch of characters (ß, Ǆ for instance) that are uppercased differently between JS and erlang. For now I blacklisted everything that doesn't work as in erlang. 

It is also possible to ignore them by returning the JS version, but it would not be in line with the erlang implementation. In my experience it's better to be safe and let it fail. 

Current implementation is ready for review.

bartblast:
@steefmin I have a suggestion for a more robust approach to handling titlecase edge cases:

Instead of manually maintaining a blacklist, you could generate a precise lookup table by comparing Erlang and JavaScript behavior across all Unicode codepoints:

Approach:

1. Create an Elixir script that iterates through all Unicode codepoints (0x0000 to 0x10FFFF) and captures the result of `:string.titlecase/1` for each character
2. Create a JavaScript script that does the same with `toUpperCase()`
3. Compare the outputs and identify where they differ (programmatically)
4. Generate a lookup table containing only the differences

This should give you a manageable set of special cases - likely just:
- Latin digraphs (Ǆ/ǅ/ǆ, Ǉ/ǈ/ǉ, Ǌ/ǋ/ǌ, Ǳ/ǲ/ǳ)
- Common ligatures (ß, ﬁ, ﬂ, etc.)
- Armenian ligatures
- Greek characters with iota subscript

Benefits:
- Guaranteed to catch all edge cases
- Creates minimal, precise lookup table (no guessing)
- Easy to verify correctness
- Can be re-run if Unicode version changes
- Replaces error-throwing blacklist with working implementation

The lookup table will likely be ~75-150 entries total, which is very reasonable and provides complete Erlang compatibility without failing on valid inputs.

Would you be interested in pursuing this approach? If you're busy or would prefer, I'm happy to take it from here and implement this solution. Either way works for me - just let me know what you'd prefer!

steefmin:
Please continue with your plan. I don't have much time to spend coming weeks.

bartblast:
Got it, I'll take it from here then. Thanks for your help @steefmin! 💜

PR #436, opened 2025-12-01
Port :erlang.bor/2 to JS

jzwood:
## Summary by CodeRabbit

* **New Features**
  * Added a new bitwise OR operation to the Erlang module for performing bitwise operations on integer values.

* **Tests**
  * Added extensive test coverage for the new operation, validating behavior across multiple scenarios including positive/negative integers, zero values, and error handling for invalid input types across both Elixir and JavaScript environments.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
@@ -3078,28 +3078,16 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
   end
 
   describe "bor/2" do
-    test "basic" do
-      assert :erlang.bor(0, 0) == 0
-      assert :erlang.bor(0, 1) == 1
-      assert :erlang.bor(1, 0) == 1
-      assert :erlang.bor(1, 0) == 1
-      assert :erlang.bor(1, 1) == 1
-    end
-
-    test "sample" do
+    test "both arguments are positive" do
       assert :erlang.bor(4, 3) == 7
-      assert :erlang.bor(-4, 3) == -1
-      assert :erlang.bor(23, 45) == 63
-      assert :erlang.bor(13, 17) == 29
-      assert :erlang.bor(101, 93) == 125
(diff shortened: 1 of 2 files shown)


test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -2388,6 +2312,44 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "bor/2" do
+    test "basic" do
```
bartblast: Could you split the tests into "basic" and "sample" categories, with each assert in a separate test? Also, it looks like some asserts are duplicates or redundant - we just need a few tests with names that clearly communicate their intent.

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -2388,6 +2312,44 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+    test "sample" do
+      assert :erlang.bor(4, 3) == 7
+      assert :erlang.bor(-4, 3) == -1
+      assert :erlang.bor(23, 45) == 63
+      assert :erlang.bor(13, 17) == 29
+      assert :erlang.bor(101, 93) == 125
+      assert :erlang.bor(33, 349) == 381
+    end
+
+    test "bor with 0" do
```
bartblast: We just need one test for when 0 is the first argument, and another for when 0 is the second argument.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -1624,6 +1624,53 @@ describe("Erlang", () => {
     });
   });
 
+  describe("bor/2", () => {
```
bartblast, line 1632: The comments for the Elixir tests apply here too. Also, please remember to make the JS and Elixir tests identical, down to the test names and values used.

bartblast:
Thank you @jzwood! 💜

PR #456, opened 2025-12-13
Port :erlang.floor/1 to JS

eburza:


bartblast:
Thank you for help @eburza! ❤️

PR #458, opened 2025-12-15
Port :erlang.binary_to_term/1 to JS

### LINKED ISSUE #457: Port :erlang.binary_to_term/1 to JS
https://hologram.page/reference/client-runtime/erlang/erlang/binary_to_term/1

0bvim:
Closes #457

Implements `:erlang.binary_to_term/1` to decode Erlang External Term Format (ETF) binaries, enabling support for `Code.fetch_docs/1` in the browser.

### Implementation

Created a dedicated `EtfDecoder` class that handles ETF parsing with support for the following tag types:

**Integers:**
- SMALL_INTEGER_EXT (97), INTEGER_EXT (98)
- SMALL_BIG_EXT (110), LARGE_BIG_EXT (111)

**Atoms:**
- ATOM_EXT (100), SMALL_ATOM_EXT (115)
- ATOM_UTF8_EXT (118), SMALL_ATOM_UTF8_EXT (119)

**Collections:**
- BINARY_EXT (109) - binaries/strings
- SMALL_TUPLE_EXT (104), LARGE_TUPLE_EXT (105)
- NIL_EXT (106), STRING_EXT (107), LIST_EXT (108)
- MAP_EXT (116)

The decoder uses a recursive approach with offset tracking, leveraging `DataView` for efficient multi-byte value reading. Error handling matches Erlang's exact error messages.

### Testing

- **JavaScript unit tests:** 23 tests covering all implemented ETF types, edge cases, and error conditions
- **Elixir consistency tests:** 26 tests verifying behavior matches OTP exactly
- All existing tests (544) continue to pass

`describetag :binary` added to `binary_to_term` test. To allow the same use of `moduletag :consistency`. But focused to execute only specific kinds of test instead all tests in module.
Later we can improve test sections to gather all tests from the same type. 
If you think that it's a good idea, later I can add it to each section in a separated `pr`.

I guess that is a good idea add a `Makefile` to this repository and add targets to ease tests and etc... what you think about?

image

image

image

## Summary by CodeRabbit

* **New Features**
  * Public Erlang binary decoder supporting full ETF term set, compressed input, and stricter validation/error messages.

* **Tests**
  * Large JS and Elixir test suites covering decoding, round-trip cases, compressed data, edge cases, and many error scenarios.

* **Tests (Helpers)**
  * Added an async test helper to assert expected boxed errors and centralized error-validation for test assertions.


assets/js/erlang/erlang.mjs
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 import Bitstring from "../bitstring.mjs";
+import EtfDecoder from "./etf_decoder.mjs";
```
bartblast: The current implementation won't work because:

This:
`import EtfDecoder from "./etf_decoder.mjs";`
will be gnored by the Hologram compiler.

The Hologram compiler only extracts code between `// Start` and `// End` delimiters. Imports at the file top are NOT bundled (they are included for tests only). The extracted code gets only the global dependencies (Type, Bitstring, Interpreter) passed as closure parameters.

**Why This Architecture?...**

Hologram uses this extraction approach to only bundle functions that are actually used by each page, keeping bundle sizes minimal. Since `binary_to_term/1` is rarely used, most applications won't need this ~300 lines of ETF decoding logic in their bundles at all. By requiring all code to be between delimiters, the compiler can perform precise tree-shaking at the function level.

**Required Fix**

Please Inline all ETF decoding logic between the delimiters as nested helper functions:

```js
// Start binary_to_term/1
"binary_to_term/1": (binary) => {
  if (!Type.isBinary(binary)) {
      Interpreter.buildArgumentErrorMsg(1, "not a binary"),
    );
  }

  // ETF tag constants
  const SMALL_INTEGER_EXT = 97;
  const INTEGER_EXT = 98;
  // ... all other constants ...

  const decodeTerm = (dataView, bytes, offset) => {
    // ... all decoding logic ...
  };

  const decodeSmallInteger = (dataView, offset) => {
    // ... helper logic ...
  };

  // ... ALL other helper functions inlined here ...

  // Main logic
  Bitstring.maybeSetBytesFromText(binary);
  const bytes = binary.bytes;
  const dataView = new DataView(bytes.buffer, bytes.byteOffset, bytes.byteLength);

  if (dataView.getUint8(0) !== 131) {
      Interpreter.buildArgumentErrorMsg(1, "invalid external representation of a term"),
    );
  }

  return decodeTerm(dataView, bytes, 1).term;
},
// End binary_to_term/1
// Deps: []
```

**Note on Testing**

Since `binary_to_term/1` currently just delegates to `EtfDecoder.decode()`, this refactoring won't affect your test complexity at all. The tests already exercise the public API (`binary_to_term/1`), so they'll continue to work exactly as-is once the code is inlined.

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1724,6 +1724,154 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "binary_to_term/1" do
+    @describetag :binary
```
bartblast: We don't need this.

test/javascript/erlang/etf_decoder_test.mjs
```diff
@@ -0,0 +1,283 @@
...
+  defineGlobalErlangAndElixirModules,
+} from "../support/helpers.mjs";
+
+import Bitstring from "../../../assets/js/bitstring.mjs";
+import Erlang from "../../../assets/js/erlang/erlang.mjs";
+import Type from "../../../assets/js/type.mjs";
+
+defineGlobalErlangAndElixirModules();
+
+describe("binary_to_term/1", () => {
```
bartblast: These tests should go to `test/javascript/erlang/erlang_test.mjs `

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1724,6 +1724,154 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "binary_to_term/1" do
```
bartblast, line 1949: It looks like there several supported ETF tags lacking dedicated tests:
- LARGE_BIG_EXT (tag 111): Implemented but not tested
- ATOM_EXT (tag 100): Implemented but not tested (only small variant tested)
- SMALL_ATOM_EXT (tag 115): Non-UTF8 variant not tested
- ATOM_UTF8_EXT (tag 118): Non-small variant not tested
- LARGE_TUPLE_EXT (tag 105): Implemented but not tested
- Improper lists: Code handles them but no test exists

Could you double check and implement what's missing?

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1724,6 +1724,154 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+
+    test "raises ArgumentError if binary is truncated" do
+      # Only version byte, no data
+      binary = <<131>>
+
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "invalid external representation of a term"),
+                   {:erlang, :binary_to_term, [binary]}
+    end
+  end
```
bartblast, line 2494: Could you add more test cases for error conditions?

for example:
- Test for unsupported ETF tag (default case in switch at etf_decoder.mjs:110-113)
- Boundary condition tests (e.g., extremely large values, malformed length fields)

etc.

assets/js/erlang/etf_decoder.mjs
```diff
@@ -0,0 +1,321 @@
+"use strict";
+
+import Bitstring from "../bitstring.mjs";
+import Interpreter from "../interpreter.mjs";
+import Type from "../type.mjs";
+
+// Erlang External Term Format (ETF) Decoder
```
bartblast: According to the https://www.erlang.org/doc/apps/erts/erl_ext_dist.html, the current implementation supports 15 out of ~30 ETF tags. The implementation is missing a few important term types that could appear in serialized Elixir/Erlang data:
- NEW_FLOAT_EXT (70) - Modern float format (8-byte IEEE floats)
- BIT_BINARY_EXT (77) - Non-byte-aligned bitstrings (e.g., <<1::3>>)
- EXPORT_EXT (113) - Captured function references (e.g., &Module.function/1)

We can skip deprecated/old formats and those related to anonymous functions, PIDs, and ports for now.

bartblast (changes requested):
Thank you for the PR! Added a few comments.

bartblast:
> I guess that is a good idea add a `Makefile` to this repository and add targets to ease tests and etc... what you think about?

Thanks for the suggestion! There are already a few mix tasks and test tags that provide similar functionality. To be honest, I don't see a strong need for such granular control of specific subsets of tests right now - you can achieve the same goal by:
- Running tests at a specific line location: `mix test path/to/test.exs:123`
- Using the @moduletag or @tag filters in Elixir tests
- Applying `.only` to a specific describe block in JS tests

If you find specific workflows that aren't well-supported by the existing tooling, we could revisit this. For now, let's keep the build tooling simple.

bartblast:
Could you pull the new changes from the dev branch? (the CI reports the PR branch has conflicts)

0bvim:
> Could you pull the new changes from the dev branch? (the CI reports the PR branch has conflicts)

ofc!

0bvim:
> > I guess that is a good idea add a `Makefile` to this repository and add targets to ease tests and etc... what you think about?
> 
> There are already a few mix tasks and test tags with similar function. To be honest I don't see a point in such a granular control of a specific subset of sets - you can achieve the same goal with running a test with specific line location in mix test or apply .only to the given describe block in JS tests.
> 
> Thanks for the suggestion! There are already a few mix tasks and test tags that provide similar functionality. To be honest, I don't see a strong need for such granular control of specific subsets of tests right now - you can achieve the same goal by:
> 
> * Running tests at a specific line location: `mix test path/to/test.exs:123`
> * Using the @moduletag or @tag filters in Elixir tests
> * Applying `.only` to a specific describe block in JS tests
> 
> If you find specific workflows that aren't well-supported by the existing tooling, we could revisit this. For now, let's keep the build tooling simple.

Nice!!! I'm used too use Makefile since my first project... So I ever want to use it... So your arguments are pretty nice. I'll get used to elixir tooling.

I will fix all requests and pull new changes from dev branch and update this `pr`!

bartblast:
Hey @0bvim, just checking in - do you need any help getting this PR across the finish line?

0bvim:
> Hey @0bvim, just checking in - do you need any help getting this PR across the finish line?

Hello @bartblast! 
Probably I'll do.
Sorry for delay in responding... this week was an chaos to me... :(
Today, sunday and next week I'll work on it... 
Now I have enough time... get it? :3

bartblast:
@0bvim Do you think you’ll finish it by next Friday? I’m planning a big release with all the ported functions around January 25th.

0bvim:
> @0bvim Do you think you’ll finish it by next Friday? I’m planning a big release with all the ported functions around January 25th.
@bartblast 
Ofc! I will ask your help if have any question!

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
0bvim: @bartblast `ATOM_EXT` is [deprecated](https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#atom_ext-deprecated). Is better I keep adding test or remove implementation and don't add test?

bartblast: @0bvim Hmm... Even though `ATOM_EXT` is deprecated in newer OTP versions, I think it's better to keep the implementation and add tests for it. Here's why:

1. **Legacy data compatibility** - Deprecated doesn't mean removed. Old saved binary data (e.g., from disk, databases, or cached ETF blobs) may still contain `ATOM_EXT` encoded atoms. We want `binary_to_term/1` to handle these correctly.

2. **OTP backward compatibility** - Systems running older Erlang/OTP versions may still produce these formats, and our decoder should handle them gracefully.

3. **Completeness** - Having full coverage of the ETF spec (including deprecated tags) makes the implementation more robust and future-proof.

So yes, please keep the implementation and add tests for it.

0bvim: @bartblast I added missing tests that you said and double checked missing term that I've seen in [Erlang Documentation](https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#introduction)

bartblast: Thank you @0bvim! Could you fix the CI failure and make sure only the new function implementation and its tests are included in the PR? I can see a few additional files, which may be related to a code formatting issue on your side (try running `mix f`).

0bvim:
I'm waiting CodeRabbit finish run and fix errors.
Waiting for your next review.

assets/js/erlang/erlang.mjs
```diff
@@ -557,6 +557,646 @@ const Erlang = {
...
+      currentOffset += 1;
+
+      const idWords = [];
+      for (let i = 0; i < len; i++) {
+        idWords.push(dataView.getUint32(currentOffset));
+        currentOffset += 4;
+      }
+
+      return {
+        term: Type.reference(nodeResult.term, creation, idWords),
```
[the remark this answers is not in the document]
0bvim: this is not valid... reference wants to receive a object and not a string

assets/js/erlang/erlang.mjs
```diff
@@ -572,6 +557,646 @@ const Erlang = {
...
+
+        case V4_PORT_EXT:
+          return decodeV4Port(dataView, bytes, offset + 1);
+
+        case EXPORT_EXT:
+          return decodeExport(dataView, bytes, offset + 1);
+
+        default:
+          Interpreter.raiseArgumentError(
+            `unsupported external term format tag: ${tag}`,
```
bartblast: The error message should match Erlang's exact error. Please use:
```js
Interpreter.buildArgumentErrorMsg(1, "invalid external representation of a term")
```
instead of the custom message. This maintains consistency with OTP behavior.

assets/js/erlang/erlang.mjs
```diff
@@ -572,6 +557,646 @@ const Erlang = {
...
+    const decodeAtom = (dataView, bytes, offset, isUtf8) => {
+      const length = dataView.getUint16(offset);
+      if (offset + 2 + length > bytes.length) {
+        Interpreter.raiseArgumentError(
+          `atom length exceeds available bytes: ${length}`,
+        );
+      }
+      const atomBytes = bytes.slice(offset + 2, offset + 2 + length);
+
+      const decoder = new TextDecoder(isUtf8 ? "utf-8" : "latin1");
```
bartblast: TextDecoder with `"latin1"` uses windows-1252, not ISO-8859-1. These encodings differ in bytes 128-159, which would cause incorrect decoding of old ATOM_EXT/SMALL_ATOM_EXT data containing those bytes.

Note: `"iso-8859-1"` won't help either - JavaScript's TextDecoder maps both to windows-1252.

For correct ISO-8859-1 decoding in `decodeAtom` and `decodeSmallAtom`, replace:

```javascript
const decoder = new TextDecoder(isUtf8 ? "utf-8" : "latin1");
const atomString = decoder.decode(atomBytes);
```

with:

```javascript
const atomString = isUtf8
  ? new TextDecoder("utf-8").decode(atomBytes)
  : String.fromCharCode(...atomBytes);
```

assets/js/erlang/erlang.mjs
```diff
@@ -572,6 +557,646 @@ const Erlang = {
...
 
+  // Start binary_to_term/1
+  "binary_to_term/1": (binary) => {
+    if (!Type.isBinary(binary)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(1, "not a binary"),
+      );
+    }
+
+    // ETF tag constants
```
bartblast, line 610: Please add support for compressed terms (tag 80 - COMPRESSED).
The format is: `80 | UncompressedSize(4 bytes) | ZlibCompressedData`.

assets/js/erlang/erlang.mjs
```diff
@@ -572,6 +557,646 @@ const Erlang = {
...
+      }
+
+      // Otherwise, it's an improper list
+      elements.push(tailResult.term);
+      return {
+        term: Type.improperList(elements),
+        newOffset: currentOffset,
+      };
+    };
+    // Map decoder
```
bartblast, line 1090: Forgot to add a blank line.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2651,834 @@ describe("Erlang", () => {
...
+        const binary = Bitstring.fromBytes(
+          new Uint8Array([
+            131, 119, 9, 116, 101, 115, 116, 95, 97, 116, 111, 109,
+          ]),
+        );
+        const result = binary_to_term(binary);
+        assert.deepStrictEqual(result, Type.atom("test_atom"));
+      });
+
+      it("decodes boolean atoms", () => {
```
bartblast: Coud you split this into two separate tests: "decodes true atom" and "decodes false atom". This follows the pattern of other tests in this file where each test covers one specific case.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2651,834 @@ describe("Erlang", () => {
...
+      it("decodes large negative integer (SMALL_BIG_EXT)", () => {
+        // :erlang.term_to_binary(-1000000000000) = <<131, 110, 5, 1, 0, 16, 165, 212, 232>>
+        const binary = Bitstring.fromBytes(
+          new Uint8Array([131, 110, 5, 1, 0, 16, 165, 212, 232]),
+        );
+        const result = binary_to_term(binary);
+        assert.deepStrictEqual(result, Type.integer(-1000000000000n));
+      });
+
+      it("decodes large positive integer (LARGE_BIG_EXT)", () => {
```
bartblast: The comment acknowledges this, but for true consistency with the Elixir test (which uses `2^2048`), consider using a value that actually requires >255 bytes. The current test manually constructs a `LARGE_BIG_EXT` binary but uses a value (1,000,000,000,000) that Erlang would encode as `SMALL_BIG_EXT`.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2651,834 @@ describe("Erlang", () => {
...
+          result,
+          Type.map([
+            [Type.atom("a"), Type.integer(1)],
+            [Type.atom("b"), Type.integer(2)],
+          ]),
+        );
+      });
+    });
+
+    describe("floats", () => {
```
bartblast, line 3090: Missing test for `FLOAT_EXT` (tag 99) - the deprecated 31-byte string format. The implementation has `decodeFloatExt` but there's no test coverage for it.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2651,834 @@ describe("Erlang", () => {
...
+        );
+      });
+
+      it("raises ArgumentError for unsupported ETF tag (tag 50)", () => {
+        // Tag 50 is not a valid ETF tag
+        const binary = Bitstring.fromBytes(new Uint8Array([131, 50]));
+        assertBoxedError(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "unsupported external term format tag: 50",
```
bartblast: If the implementation is updated to use Erlang's standard error message (`"invalid external representation of a term"`), these tests will need to be updated accordingly.

mix.exs
```diff
@@ -11,9 +11,9 @@ defmodule Hologram.MixProject do
     [
       eslint:
         "cmd assets/node_modules/.bin/eslint --color --config assets/eslint.config.mjs assets/js/** benchmarks/javascript/** scripts/** test/javascript/** --no-error-on-unmatched-pattern",
-      f: ["format", "format.js", "cmd cd test/features && mix format"],
+      f: ["format", "format.js", "cmd --cd test/features mix format"],
```
bartblast: These changes (`format` command updates, `preferred_cli_env` -> `cli` refactor) are unrelated to `binary_to_term/1` and should be submitted as a separate PR per the contribution guidelines ("one function per pull request").

assets/package-lock.json
```diff
@@ -298,7 +298,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
```
bartblast, line 301: This change (removing `"peer"` flags) is unrelated to `binary_to_term/1` - could you revert it. This was probably caused by using a different Node.js version than the one specified in `.tool-versions`. Please ensure you're using the correct Node.js version (via asdf or similar) before running `npm install`.

test/elixir/support/fixtures/compiler/transformer/module_109.ex
```diff
@@ -4,7 +4,7 @@ defmodule Hologram.Test.Fixtures.Compiler.Tranformer.Module109 do
 
   defstruct [:a, :b]
 
-  def test(x) do
+  def test(%Module109{} = x) do
```
bartblast: This change (adding pattern match `%Module109{}` to function signature) appears unrelated to `binary_to_term/1`. Could you explain why this change is needed? If it's a separate fix, it should be submitted as a separate PR.

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1946,6 +1891,309 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
     end
   end
 
+  describe "binary_to_term/1" do
```
bartblast, line 1949: Could you a test for `FLOAT_EXT` (tag 99) - the deprecated 31-byte string format. Example:

```elixir
test "decodes FLOAT_EXT (deprecated string format)" do
  # Manually construct FLOAT_EXT binary
  float_str = :io_lib.format("~.20e", [1.5]) |> IO.iodata_to_binary()
  padded = String.pad_trailing(float_str, 31, <<0>>)
  binary = <<131, 99, padded::binary>>

  assert :erlang.binary_to_term(binary) == 1.5
end
```

bartblast (changes requested):
Thank you! :) Added some comments.

assets/js/erlang/erlang.mjs
0bvim: I added, but I used a lib to compress and decompress. I guess that I need to make this functionality instead use a lib  (pako), right? @bartblast

bartblast: @0bvim Why not Compression Streams API?

0bvim: Oh... I didn't even know that it exists... I'll use it. Ty

bartblast:
Hey @0bvim, how's it going with this PR? Any update on the progress? Need any help?

0bvim:
> Hey @0bvim, how's it going with this PR? Any update on the progress? Need any help?

Hey @bartblast I guess that I'll do all the fix today. 
All progress that I've made is up to date with this branch. 
This week was very busy.

bartblast:
Thanks for the update @0bvim! Could you run the formatter: `mix f` when you get a chance?

0bvim:
> Thanks for the update @0bvim! Could you run the formatter: `mix f` when you get a chance?

Done @bartblast

bartblast:
Hmm... that's odd - CI still shows formatting problems.

0bvim:
> Hmm... that's odd - CI still shows formatting problems.

@bartblast let me fix all the issues, revert needed content and format again and we see if it still happening.

assets/js/erlang/erlang.mjs
0bvim: @bartblast hey, this guy take a ruge refactor due Zlib needs to be async. 
I fix almost all things, now I just need to fix missing tests under `erlang_test.mjs`. 
I'll do it "tomorrow". Now is too late here and I really need to sleep xD

test/elixir/support/fixtures/compiler/transformer/module_109.ex
0bvim: It appear in warnings and I fixed a lot of this things because of new version of Elixir. But I reverted most part and forgot about this one.

assets/js/erlang/erlang.mjs
bartblast: Thanks @0bvim for tackling this, I know async refactors can be a pain! 🙏
One request: could you move all the decompression-related code from the Utils class directly into the ported function itself? The reason is that we're aggressively optimizing bundle size through call graph analysis - if it lives in Utils, it'll get included in every Hologram application even if they never use this function. By keeping it inside the ported function, the Hologram compiler can tree-shake it out when it's not needed.

0bvim: Np @bartblast! I want to contribute with more later! 

I'll move it, of course! 
I'll pay more attention to this question of bundle size.
How I can check this graph analysis? You can send some link or taught me how to?

bartblast: It's not really documented yet.

0bvim:
@bartblast I can stack a `pr` fixing warning from `mix.exs` about use `cli()` instead of `preferred_cli_env()` into this `pr`? 
Or is better I open a brand new `pr` to `dev` and then rebase in this `pr`?

bartblast:
@0bvim I'd like to keep the PR focused on the ported function only.

bartblast:
@0bvim Could you address CodeRabbit's code review first (there are some unresolved comments)?

0bvim:
> @0bvim Could you address CodeRabbit's code review first (there are some unresolved comments)?

@bartblast I've solved all now!

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+
+      it("raises ArgumentError for compressed term with truncated uncompressed size", async () => {
+        // COMPRESSED tag but missing uncompressed size bytes
+        const binary = Bitstring.fromBytes(
+          new Uint8Array([131, 80, 0, 0]), // Only 2 bytes of size instead of 4
+        );
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "errors were found at the given arguments:\n\n  * 1st argument: invalid external representation of a term\n",
```
bartblast: Could you use `Interprerter.buildArgumentErrorMsg()` helper here and in similar error case tests?

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+          new Uint8Array([131, 109, 0, 0, 0, 10, 65, 66]),
+        );
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "binary length exceeds available bytes: 10",
+        );
+      });
+
+      it("raises ArgumentError for malformed STRING_EXT with length exceeding data", async () => {
```
bartblast, line 3756: There's no corresponding Elixir consistency test.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+          new Uint8Array([131, 116, 0, 0, 0, 10]),
+        );
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "errors were found at the given arguments:\n\n  * 1st argument: invalid external representation of a term\n",
+        );
+      });
+
+      it("raises ArgumentError for malformed ATOM_EXT with length exceeding data", async () => {
```
bartblast, line 3816: There's no corrresponding Elixir consistency test.

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+        newOffset: offset + 5 + n,
+      };
+    };
+
+    // Atom decoders
+
+    const decodeAtom = (dataView, bytes, offset, isUtf8) => {
+      const length = dataView.getUint16(offset);
+      if (offset + 2 + length > bytes.length) {
+        Interpreter.raiseArgumentError(
```
bartblast, line 937: The error message here doesn't use `Interpreter.buildArgumentErrorMsg()` like other error handlers in this function. This will produce a different error format than Erlang.

**Should be:**
```javascript
  Interpreter.buildArgumentErrorMsg(1, "invalid external representation of a term"),
);
```

(Make sure that JS and Erlang behaviour for this case matches through tests)

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+          new Uint8Array([131, 100, 0, 100, 65, 66]),
+        );
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "invalid external representation of a term",
+        );
+      });
+
+      it("raises ArgumentError for malformed SMALL_ATOM_UTF8_EXT with length exceeding data", async () => {
```
bartblast, line 3831: There's no corresponding Elixir test to test that the behaviour is exactly the same for Elixir and JS.

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+        term: Type.atom(atomString),
+        newOffset: offset + 2 + length,
+      };
+    };
+
+    const decodeSmallAtom = (dataView, bytes, offset, isUtf8) => {
+      const length = dataView.getUint8(offset);
+      if (offset + 1 + length > bytes.length) {
+        Interpreter.raiseArgumentError(
+          `atom length exceeds available bytes: ${length}`,
```
bartblast: This may be a custom error message, but we need to match Erlang behaviour.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+        // Tag 1 is not a valid ETF tag
+        const binary = Bitstring.fromBytes(new Uint8Array([131, 1]));
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "errors were found at the given arguments:\n\n  * 1st argument: invalid external representation of a term\n",
+        );
+      });
+
+      it("raises ArgumentError for malformed BINARY_EXT with length exceeding data", async () => {
```
bartblast, line 3741: There's no corresponding Elixir consistency test.

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+      };
+    };
+
+    // Binary decoder
+
+    const decodeBinary = (dataView, bytes, offset) => {
+      const length = dataView.getUint32(offset);
+      if (offset + 4 + length > bytes.length) {
+        Interpreter.raiseArgumentError(
+          `binary length exceeds available bytes: ${length}`,
```
bartblast: This may be a custom error message, but we need to match the Erlang behaviour here.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+          new Uint8Array([131, 111, 0, 0, 1, 0, 0, 1, 2, 3]),
+        );
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "errors were found at the given arguments:\n\n  * 1st argument: invalid external representation of a term\n",
+        );
+      });
+
+      it("raises ArgumentError for malformed BIT_BINARY_EXT with length exceeding data", async () => {
```
bartblast, line 3876: There's no corresponding Elixir consistency test.

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+          Interpreter.buildArgumentErrorMsg(
+            1,
+            "invalid external representation of a term",
+          ),
+        );
+      }
+
+      if (offset + 5 + length > bytes.length) {
+        Interpreter.raiseArgumentError(
+          `bit binary length exceeds available bytes: ${length}`,
```
bartblast: Same issue here - custom error message doesn't match Erlang's output.

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+        bytes[1] = 99;
+        for (let i = 0; i < floatStr.length; i++) {
+          bytes[2 + i] = floatStr.charCodeAt(i);
+        }
+        const binary = Bitstring.fromBytes(bytes);
+        const result = await binary_to_term(binary);
+        assert.deepStrictEqual(result, Type.float(0.00625));
+      });
+
+      it("raises ArgumentError for FLOAT_EXT with truncated data", async () => {
```
bartblast, line 3192: No corresponding Elixir consistency test.

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+      return {
+        term: Type.float(value),
+        newOffset: offset + 8,
+      };
+    };
+
+    const decodeFloatExt = (dataView, bytes, offset) => {
+      // FLOAT_EXT: 31-byte null-terminated string (deprecated format)
+      if (offset + 31 > bytes.length) {
+        Interpreter.raiseArgumentError("float string exceeds available bytes");
```
bartblast: Same issue - custom error message doesn't match Erlang's output.

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+        const result = await decodeTerm(decompressedView, decompressed, 0);
+
+        // Return with newOffset pointing to end of original bytes since
+        // COMPRESSED must consume the entire remaining buffer at top-level
+        return {
+          term: result.term,
+          newOffset: bytes.length,
+        };
+      } catch (error) {
+        console.error("Decode error:", error);
```
bartblast: Could you remove this line?

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+      if (offset + 5 + length > bytes.length) {
+        Interpreter.raiseArgumentError(
+          `bit binary length exceeds available bytes: ${length}`,
+        );
+      }
+
+      const binaryBytes = bytes.slice(offset + 5, offset + 5 + length);
+      const bitstring = Bitstring.fromBytes(new Uint8Array(binaryBytes));
+
+      // Adjust leftoverBitCount based on the bits field
```
bartblast, line 1182: The comment says "If bits is 0, all bytes are full" but the validation above rejects bits=0 with `if (bits < 1 || bits > 8)`. According to ETF spec, BIT_BINARY_EXT's Bits field should be 1-8, so the validation is correct but the comment is misleading.

**Should be:**
```javascript
// Adjust leftoverBitCount based on the bits field
// If bits is 8, all bytes are full (leftoverBitCount = 0)
// If bits is 1-7, the last byte is partial (leftoverBitCount = bits)
```

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+        );
+        await assertBoxedErrorAsync(
+          () => binary_to_term(binary),
+          "ArgumentError",
+          "errors were found at the given arguments:\n\n  * 1st argument: invalid external representation of a term\n",
+        );
+      });
+    });
+
+    describe("error handling", () => {
```
bartblast, line 3659: Once the error messages are fixed to match Erlang's format (using `buildArgumentErrorMsg`), the following tests will need their expected messages updated:

1. `"raises ArgumentError for malformed BINARY_EXT with length exceeding data"` - currently expects `"binary length exceeds available bytes: 10"`
2. `"raises ArgumentError for malformed SMALL_ATOM_UTF8_EXT with length exceeding data"` - currently expects `"atom length exceeds available bytes: 50"`
3. `"raises ArgumentError for malformed BIT_BINARY_EXT with length exceeding data"` - currently expects `"bit binary length exceeds available bytes: 100"`
4. `"raises ArgumentError for FLOAT_EXT with truncated data"` - currently expects `"float string exceeds available bytes"`

etc.

(Remember to write matching Elixir consistency tests)

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+            103, 202, 207, 78, 212, 26, 165, 7, 7, 157, 5, 0, 189, 136, 115, 25,
+          ]),
+        );
+        const result = await binary_to_term(binary);
+        const expectedTuple = Type.tuple([Type.atom("ok"), Type.integer(42)]);
+        const expectedList = Type.list(Array(50).fill(expectedTuple));
+        assertBoxedStrictEqual(result, expectedList);
+      });
+
+      it("validates internal zlibInflate functionality within binary_to_term context", async () => {
```
bartblast: This test is redundant. It uses the same input data and assertions as `"decodes compressed term (COMPRESSED tag 80) with repeated string"

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -1946,6 +1946,379 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
...
+                   {:erlang, :binary_to_term, [binary]}
+    end
+
+    test "raises ArgumentError if argument is not a binary" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(1, "not a binary"),
+                   {:erlang, :binary_to_term, [:test]}
+    end
+
+    test "raises ArgumentError if argument is a non-binary bitstring" do
```
bartblast, line 2313: There's no corresponding JS test, please add:
```javascript
it("raises ArgumentError if argument is a non-binary bitstring", async () => {
  const bits = Bitstring.fromBits([1, 0, 1]);

  await assertBoxedErrorAsync(
    () => binary_to_term(bits),
    "ArgumentError",
    Interpreter.buildArgumentErrorMsg(1, "not a binary")
  );
});
```

test/javascript/erlang/erlang_test.mjs
```diff
@@ -2753,6 +2754,1053 @@ describe("Erlang", () => {
...
+          (arity >> 8) & 0xff,
+          arity & 0xff,
+        ];
+        const header = new Uint8Array([131, 105, ...arityBytes]);
+
+        // Create element bytes: each element is SMALL_INTEGER_EXT (tag 97) + value
+        // Values 1 to 300, but modulo 256 for values > 255
+        const elementBytes = [];
+        for (let i = 1; i <= arity; i++) {
+          elementBytes.push(97, i & 0xff); // SMALL_INTEGER_EXT + value (wraps at 256)
```
bartblast: The test creates element values with `i & 0xff`, which means values wrap at 256 (e.g., element 256 has value 0, element 257 has value 1). This is fine for testing the decoder, but a comment explaining this would be helpful for future maintainers.

**Action:** Add a comment:
```javascript
// Values 1 to 300, but SMALL_INTEGER_EXT only supports 0-255,
// so we use modulo 256 to keep all elements as SMALL_INTEGER_EXT
elementBytes.push(97, i & 0xff);
```

test/javascript/support/helpers.mjs
```diff
@@ -93,6 +93,57 @@ export function assertBoxedError(
   }
 }
 
+export async function assertBoxedErrorAsync(
```
bartblast, line 96: The new `assertBoxedErrorAsync` function duplicates ~50 lines of validation logic from the existing `assertBoxedError`. The only difference is `await asyncCallable()` vs `callable()`.

Please extract the common validation logic into a shared helper, e.g.:

```javascript
function validateBoxedError(error, expectedErrorType, expectedErrorMessage) {
  const isRegex = expectedErrorMessage instanceof RegExp;
  const expectedMessageDisplay = isRegex
    ? expectedErrorMessage.toString()
    : expectedErrorMessage;

  const failMessagePrefix = `\nexpected:\n${expectedErrorType}: ${expectedMessageDisplay}\n`;

  if (!error) {
    assert.fail(failMessagePrefix + "but got no error");
  }

  if (!(error instanceof HologramBoxedError)) {
    assert.fail(
      failMessagePrefix + `but got:\n${error.name}: ${error.message}`,
    );
  }

  const receivedErrorType = Interpreter.getErrorType(error);
  const receivedErrorMessage = Interpreter.getErrorMessage(error);

  const typeMatches = receivedErrorType === expectedErrorType;
  const messageMatches = isRegex
    ? expectedErrorMessage.test(receivedErrorMessage)
    : Interpreter.isStrictlyEqual(
        error.struct,
        Type.errorStruct(expectedErrorType, expectedErrorMessage),
      );

  if (!typeMatches || !messageMatches) {
    assert.fail(
      failMessagePrefix +
        `but got:\n${receivedErrorType}: ${receivedErrorMessage}`,
    );
  }
}

export function assertBoxedError(callable, expectedErrorType, expectedErrorMessage) {
  let error;
  try { callable(); } catch (e) { error = e; }
  validateBoxedError(error, expectedErrorType, expectedErrorMessage);
}

export async function assertBoxedErrorAsync(asyncCallable, expectedErrorType, expectedErrorMessage) {
  let error;
  try { await asyncCallable(); } catch (e) { error = e; }
  validateBoxedError(error, expectedErrorType, expectedErrorMessage);
}
```

assets/js/erlang/erlang.mjs
```diff
@@ -599,6 +599,782 @@ const Erlang = {
...
+          Interpreter.buildArgumentErrorMsg(
+            1,
+            "invalid external representation of a term",
+          ),
+        );
+      }
+
+      const tag = dataView.getUint8(offset);
+
+      switch (tag) {
```
bartblast, line 696: I noticed the following ETF tags are not supported in the implementation:

| Tag | Code | Purpose |
|-----|------|---------|
| REFERENCE_EXT | 101 | Old reference format (backwards compatibility) |
| NEW_FUN_EXT | 112 | Closures with captured variables |
| FUN_EXT | 117 | Old closure format (backwards compatibility) |

While `binary_to_term/1` can encounter any of these in valid ETF binaries, in practice for Hologram we should only need **REFERENCE_EXT** right now.

Please add support for **REFERENCE_EXT (tag 101)**.

For the other missing tags (NEW_FUN_EXT, FUN_EXT), I'll make a note for possible implementation in the future.

bartblast (changes requested):
Thanks for the work on this - it's a substantial implementation that covers a lot of ground.

I've left some comments on specific issues. The main areas are:
- Inconsistent error messages (some use custom messages instead of `buildArgumentErrorMsg`)
- `console.error` left in production code
- Missing REFERENCE_EXT support
- Some test coverage gaps

I didn't cover everything in detail - please make sure the behaviour is fully consistent between the JS and Elixir implementations. The best way to achieve this is to ensure the JS tests and Elixir consistency tests are **exactly mirrored**: same test names, same order, same arguments. This makes it much easier to verify correctness and maintain going forward.

assets/js/erlang/erlang.mjs
0bvim: Done

bartblast:
Thanks for the work on this, @0bvim!

Given the size of the PR, I'd like to push merging this to the next minor release in 1-2 weeks so I can review it thoroughly. I'll get back to you soon!

bartblast:
Hey @0bvim, just a heads up - since `binary_to_term/1` is an async function, we need async function calling support in the Hologram runtime before this can be merged. I'm currently working on exactly that as part of JS interop, and it should be released very soon. Once that's in place, this PR will be unblocked and I'll get to the review. Thanks for your patience!

0bvim:
> Hey @0bvim, just a heads up - since `binary_to_term/1` is an async function, we need async function calling support in the Hologram runtime before this can be merged. I'm currently working on exactly that as part of JS interop, and it should be released very soon. Once that's in place, this PR will be unblocked and I'll get to the review. Thanks for your patience!

Nice @bartblast, thanks for let me know!

bartblast:
Thank you @0bvim! 💜 🎉

0bvim:
> Thank you @0bvim! 💜 🎉

You're welcome! 
I've got a new job now and I'm busy learning new stuff and so on. But I want to contribute more later. 
Congrats for sponsor and partnership with curiosum, well deserved man!

PR #482, opened 2025-12-28
Use the last formatting option in :erlang.float_to_binary/2

### LINKED ISSUE #478: Use the last formatting option in :erlang.float_to_binary/2
:erlang.float_to_binary(7.12, [{:decimals, 4}, {:scientific, 3}]) 

Expected  "7.120e+00"
Actual  "7.1200"

ideaMarcos:
Closes #478 

## Summary by CodeRabbit

* **Bug Fixes**
  * float_to_binary/2 now consistently honors only the last-specified format option when multiple formats (short, decimals, scientific) are provided, ensuring deterministic formatting.

* **Tests**
  * Expanded test coverage to verify last-format precedence across implementations and option orderings, including combos with compact/short/decimals/scientific and relevant edge cases.


test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
```diff
@@ -2494,6 +2494,18 @@ defmodule Hologram.ExJsConsistency.Erlang.ErlangTest do
         :erlang.float_to_binary(@input_above_10, [:abc])
       end
     end
+
+    test "multiple formats - use last format {:scientific}" do
```
bartblast: Could you add a few tests to demonstrate that it works correctly when `:compact` option is used (on different positions and with different other options)?

bartblast (changes requested):

bartblast:
@coderabbitai full review

test/elixir/hologram/ex_js_consistency/erlang/erlang_test.exs
ideaMarcos: sure. there are already many tests for compact around line 2410 too

bartblast: Yep, but they're only covering the decimals case.

bartblast: PS: I realised I wrote "Could I add" in the initial comment - corrected to "Could you add".

ideaMarcos: https://www.erlang.org/doc/apps/erts/erlang.html#float_to_list/2

`If option compact is specified, the trailing zeros at the end of the list are truncated. This option is only meaningful together with option decimals.`

so compact only works with decimals

i can add test for 

```elixir
    test ":compact option by itself is same as default format" do
      assert :erlang.float_to_binary(@input_above_10, [:compact]) ==
               :erlang.float_to_binary(@input_above_10, [])
    end
```

bartblast: Yes, but the user may pass incorrect opts and we need to verify that the client-side behaviour is consistent with the server-side behaviour in this regard.

ideaMarcos: how do these erlang tests look? if good, i'll port them to js

bartblast: Look good, please go ahead.

bartblast:
Thank you @ideaMarcos! 💜

PR #483, opened 2025-12-29
Port :maps.take/2 to JS

### LINKED ISSUE #357: Port :maps.take/2 to JS
Removes Key and its associated value from Map1, if it exists, returning a tuple
with the removed value Value and the new map Map2; otherwise, returns error.

The call will fail with a {badmap,Map} exception if Map1 is not a map.

Example:

    1> Map = #{"a" => "hello", "b" => "world"}.
    #{"a" => "hello", "b" => "world"}
    2> maps:take("a", Map).
    {"hello",#{"b" => "world"}}
    3> maps:take("does not exist", Map).
    error

franzinBr:
Closes #357 

## Summary by CodeRabbit

* **New Features**
  * Added a map operation that retrieves and removes a key-value pair in one step, returning the value and remaining map on success, or an error when the key is missing.

* **Tests**
  * Added comprehensive tests covering successful retrieval, missing keys, empty maps, and invalid-argument error handling.

### DIFF
--- a/assets/js/erlang/maps.mjs
+++ b/assets/js/erlang/maps.mjs
@@ -282,10 +282,9 @@ const Erlang_Maps = {
 
   // Start take/2
   "take/2": (key, map1) => {
-    const SENTINEL = Symbol("not_found");
-    const value = Erlang_Maps["get/3"](key, map1, SENTINEL);
+    const value = Erlang_Maps["get/3"](key, map1, null);
 
-    if (value === SENTINEL) {
+    if (value === null) {
       return Type.atom("error");
     }
 


assets/js/erlang/maps.mjs
```diff
@@ -279,6 +279,21 @@ const Erlang_Maps = {
   },
   // End values/1
   // Deps: []
+
+  // Start take/2
+  "take/2": (key, map1) => {
+    const SENTINEL = Symbol("not_found");
```
franzinBr: I'm using a `Symbol` as a sentinel value here to avoid an extra `is_key/2` lookup. Since [Symbols are guaranteed to be unique](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), there's no risk of collision with actual map values. Let me know if you'd prefer a different approach, @bartblast

bartblast: Hologram uses boxed values, so you can just use `null` as the sentinel here. No boxed value will ever be plain null, so the `Symbol` is unnecessary.

bartblast:
Hey @franzinBr, just checking in - do you need any help getting this PR across the finish line?

assets/js/erlang/maps.mjs
franzinBr: Make sense!

franzinBr:
> Hey @franzinBr, just checking in - do you need any help getting this PR across the finish line?

Change applied, sorry for the delay.

bartblast:
Thank you @franzinBr! 💜

PR #529, opened 2026-01-08
Port :maps.intersect_with/3 to JS

### LINKED ISSUE #524: Port :maps.intersect_with/3 to JS

simonmcconnell:
Closes #524

## Summary by CodeRabbit

* **New Features**
  * Added intersect_with/3 to the Erlang maps API to combine entries from two maps using a provided 3-arity function.

* **Tests**
  * Added comprehensive Elixir and JavaScript tests for intersect_with/3 covering combination behavior, multiple keys, empty-input cases, type and arity validations, and non-mutation of inputs; note: test blocks were added twice.

### DIFF
--- a/test/javascript/erlang/maps_test.mjs
+++ b/test/javascript/erlang/maps_test.mjs
@@ -360,13 +360,13 @@ describe("Erlang_Maps", () => {
       const map1 = Type.map([
         [Type.atom("a"), Type.integer(1)],
         [Type.bitstring("a"), Type.integer(2)],
-        [Type.integer("1"), Type.integer(3)],
+        [Type.integer(1), Type.integer(3)],
       ]);
 
       const map2 = Type.map([
         [Type.atom("a"), Type.integer(10)],
         [Type.bitstring("a"), Type.integer(20)],
-        [Type.integer("1"), Type.integer(30)],
+        [Type.integer(1), Type.integer(30)],
       ]);
 
       const result = intersect_with(fun, map1, map2);


assets/js/erlang/maps.mjs
```diff
@@ -111,6 +112,41 @@ const Erlang_Maps = {
...
   // Deps: []
 
+  // Start intersect_with/3
+  "intersect_with/3": (fun, map1, map2) => {
+    if (!Type.isAnonymousFunction(fun) || fun.arity !== 3) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(
+          1,
+          "not a fun that takes three arguments",
+          [],
```
bartblast: The 3rd arg is redundant.

test/elixir/hologram/ex_js_consistency/erlang/maps_test.exs
```diff
@@ -178,6 +178,43 @@ defmodule Hologram.ExJsConsistency.Erlang.MapsTest do
     end
   end
 
+  describe "intersect_with/3" do
```
bartblast, line 151: Could you order the describe block alphabetically (it should go before the `is_key/2` block)

test/elixir/hologram/ex_js_consistency/erlang/maps_test.exs
```diff
@@ -178,6 +178,43 @@ defmodule Hologram.ExJsConsistency.Erlang.MapsTest do
...
+    end
+
+    test "returns an empty map when map1 is empty" do
+      assert :maps.intersect_with(fn _k, v1, v2 -> v1 + v2 end, %{}, %{a: 2}) == %{}
+    end
+
+    test "returns an empty map when map2 is empty" do
+      assert :maps.intersect_with(fn _k, v1, v2 -> v1 + v2 end, %{a: 1}, %{}) == %{}
+    end
+
```
bartblast, line 180: There's no test for when both map1 and map2 are empty.

test/elixir/hologram/ex_js_consistency/erlang/maps_test.exs
```diff
@@ -178,6 +178,43 @@ defmodule Hologram.ExJsConsistency.Erlang.MapsTest do
...
   end
 
+  describe "intersect_with/3" do
+    test "combines values with function" do
+      assert :maps.intersect_with(fn _k, v1, v2 -> v1 + v2 end, %{a: 1, b: 3}, %{a: 2, c: 4}) ==
+               %{
+                 a: 3
+               }
+    end
+
```
bartblast, line 158: Could you add a test with multiple common keys to verify all intersecting keys are processed?

test/elixir/hologram/ex_js_consistency/erlang/maps_test.exs
```diff
@@ -178,6 +178,43 @@ defmodule Hologram.ExJsConsistency.Erlang.MapsTest do
...
+        :maps.intersect_with(fn _k, v1, v2 -> v1 + v2 end, :abc, %{})
+      end
+    end
+
+    test "raises when map2 is not a map" do
+      assert_error BadMapError, "expected a map, got: :abc", fn ->
+        :maps.intersect_with(fn _k, v1, v2 -> v1 + v2 end, %{}, :abc)
+      end
+    end
+
```
bartblast, line 192: The JS tests include "raises when function is not a function" (passing Type.atom("abc")), but there's no corresponding Elixir consistency test:

test/javascript/erlang/maps_test.mjs
```diff
@@ -319,6 +319,150 @@ describe("Erlang_Maps", () => {
     });
   });
 
+  describe("intersect_with/3", () => {
```
bartblast, line 322: Some comments for Elixir tests apply here too.

bartblast (changes requested):

assets/js/erlang/maps.mjs
```diff
@@ -111,6 +112,41 @@ const Erlang_Maps = {
...
+    }
+
+    if (!Type.isMap(map1)) {
+      Interpreter.raiseBadMapError(map1);
+    }
+    if (!Type.isMap(map2)) {
+      Interpreter.raiseBadMapError(map2);
+    }
+
+    return Type.map(
```
bartblast: The return statement is quite dense, could you refactor it a little bit for readability? E.g.:

```js
const entries = Object.values(map1.data).reduce((acc, [key, value]) => {
  const encodedKey = Type.encodeMapKey(key);
  const keyValue2 = map2.data[encodedKey];

  if (keyValue2) {
    const combinedValue = Interpreter.callAnonymousFunction(fun, [key, value, keyValue2[1]]);
    acc.push([key, combinedValue]);
  }

  return acc;
}, []);

return Type.map(entries);
```

test/elixir/hologram/ex_js_consistency/erlang/maps_test.exs
simonmcconnell: No worries.  In the erlang source they check which map is smallest and iterate over that.  I'm not sure if map_size will be as efficient on the JS side.  You want that code ported over or left iterating over the first map?

bartblast: @simonmcconnell You're right that `map_size` isn't as efficient on the JS side - we'd need `Object.keys(map.data).length` which is O(n) vs Erlang's O(1). So let's just iterate over `map1` for now.

Generally speaking, we don't need to mirror Erlang's implementation exactly - just the behaviour. The platforms have different characteristics, so we pick what works best for JS.

Could you add a brief comment explaining this decision? Something noting that Erlang iterates over the smaller map, but we skip that since getting map size in JS isn't O(1).

(Side note: tracking size on the boxed map type could be beneficial down the road - possibly when we get to structural sharing - and this implementation could be revised then. But that's for another time.)

bartblast:
Hey @simonmcconnell! Just a heads up - I'm planning to release all the ported functions in a few days. Any chance you'll be able to wrap this one up soon?

test/elixir/hologram/ex_js_consistency/erlang/maps_test.exs
simonmcconnell: What is the reason for using an Object vs Map on the JS side?

bartblast: Eventually we'll probably migrate to Map for the data field for boxed maps, but we'll make the decision only after I get into planning structural sharing implementation.

simonmcconnell: If that happens, we can change this to be the same as the Erlang implementation.

bartblast:
Thank you @simonmcconnell! 💜

PR #534, opened 2026-01-08
Refactor :filename.basename/1 to use :filename.flatten/1

### LINKED ISSUE #532: Refactor :filename.basename/1 in JS
The currently implemented version of this function should depend on `:filename.flatten/1` (see #528). Currently it implements its own functionality to do what should be done by `:filename.flatten/1`, and is probably incomplete in its implementation - though I haven't studied it in detail.

One big issue is that the guard clause in `:filename.basename/1` reports an error that purports to come from `:filename.do_flatten/2`, which is a private function used by `:filename.flatten/1`. Reporting an error as coming from a function other than the one actually throwing the error is disingenuous and misleading.

Once `:filename.flatten/1` - and by extension `:filename._do_flatten/2` - is merged then `:filename.basename/1` should be refactored to use the new function and to much more closely reflect the Erlang implementation of the function.

mward-sudo:
Closes #532 

## Summary by CodeRabbit

* **Refactor**
  * Unified filename normalization so basename extraction is consistent across input types.
* **Bug Fixes**
  * Preserve raw bytes for filename components containing invalid UTF‑8, returning binary or byte-list output rather than attempting to decode.
* **Tests**
  * Added cross-language tests to verify raw-byte preservation for invalid UTF‑8 inputs.
* **Chore**
  * Public API signature for basename remains unchanged.

### DIFF
--- a/assets/js/erlang/filename.mjs
+++ b/assets/js/erlang/filename.mjs
@@ -133,7 +133,7 @@ const Erlang_Filename = {
     };
 
     const handleListFilename = (flattened) => {
-      const {component, isValidUtf8} = getBinaryAndValidateUtf8(
+      const {component, basenameBinary, isValidUtf8} = getBinaryAndValidateUtf8(
         flattened,
         false,
       );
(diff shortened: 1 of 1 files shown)


bartblast:
@mward-sudo Could you pull new changes from dev (this should fix the CI fails).

bartblast:
Thank you @mward-sudo! 💜

PR #540, opened 2026-01-16
Port :filename.dirname/1 to JS

### LINKED ISSUE #412: Port :filename.dirname/1 to JS

mward-sudo:
Closes #412 

## Summary by CodeRabbit

* **New Features**
  * Added directory-name extraction across many path formats and input types (binaries, lists, iolists, atoms). Preserves input format and correctly handles trailing slashes, root/empty paths, dot components, and invalid UTF‑8. Public API now exposes dirname capability.

* **Tests**
  * Added extensive Elixir and JavaScript tests covering relative/absolute paths, edge cases, input types, and error scenarios.

### DIFF
--- a/assets/js/erlang/filename.mjs
+++ b/assets/js/erlang/filename.mjs
@@ -14,6 +14,13 @@ const Erlang_Filename = {
   "_dirname_raw/1": (filenameBinary) => {
     // Helpers
 
+    const trimTrailingSeparators = (bytes, DIR_SEPARATOR_BYTE) => {
+      let end = bytes.length - 1;
+      while (end >= 0 && bytes[end] === DIR_SEPARATOR_BYTE) end--;
+      if (end < 0) return [DIR_SEPARATOR_BYTE];
+      return bytes.slice(0, end + 1);
+    };
+
     const computeResultBytes = (
       lastSlashIndex,
       trimmedBytes,
@@ -29,7 +36,11 @@ const Erlang_Filename = {
 
       // Return bytes before the last separator (convert reversed index back to normal index)
       // Formula: normal_index = array_length - 1 - reversed_index
-      return trimmedBytes.slice(0, trimmedBytes.length - 1 - lastSlashIndex);
+      const raw = trimmedBytes.slice(
+        0,
+        trimmedBytes.length - 1 - lastSlashIndex,
+      );
+      return trimTrailingSeparators(raw, DIR_SEPARATOR_BYTE);
     };
 
     // Search from end of array; returns reversed index (distance from end)
@@ -66,6 +77,13 @@ const Erlang_Filename = {
     }
 
     const trimmedBytes = bytes.slice(0, bytes.length - reversedIndex);
+    const hadTrailingSeparators = reversedIndex > 0;
+
+    if (hadTrailingSeparators) {
+      const result = Bitstring.fromBytes(trimmedBytes);
+      Bitstring.maybeSetTextFromBytes(result);
+      return result;
+    }
 
     // Find last separator in trimmed bytes
     const lastSlashIndex = findLastSeparatorIndex(
--- a/test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
@@ -331,6 +331,10 @@ defmodule Hologram.ExJsConsistency.Erlang.FilenameTest do
       assert :filename.dirname("foo/bar/baz/") == "foo/bar/baz"
     end
 
+    test "multi-component path with multiple trailing slashes" do
+      assert :filename.dirname("foo/bar//") == "foo/bar"
+    end
+
     test "single dot" do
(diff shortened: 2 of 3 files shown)


assets/js/erlang/filename.mjs
```diff
@@ -52,6 +52,82 @@ const Erlang_Filename = {
   // End _do_flatten/2
   // Deps: [:erlang.atom_to_list/1]
 
+  // Start _dirname_raw/1
+  "_dirname_raw/1": (filenameBinary) => {
```
bartblast, line 14: Could you order the functions alphabetically? (`_dirname_raw/1` before `_do_flatten/2`)

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -98,6 +98,112 @@ defmodule Hologram.ExJsConsistency.Erlang.FilenameTest do
...
+
+    test "double dots" do
+      assert :filename.dirname("..") == "."
+    end
+
+    test "absolute path with multiple components" do
+      assert :filename.dirname("/foo/bar") == "/foo"
+    end
+
+    test "deep relative path" do
```
bartblast: "deep relative path" (`"a/b/c/d/e"`) is redundant with "path with multiple components" (`"foo/bar/baz.erl"`). Both exercise the same code path: find the last `/` in a relative multi-component path and return everything before it. The only difference is the number of components, which doesn't exercise any additional logic. Please remove "deep relative path" from both the JS and Elixir test files.

Also, rename "path with multiple components" to **"relative path with multiple components"** for consistency with the existing "absolute path with multiple components" test, and move "absolute path with multiple components" next to it so the two related tests are grouped together.

(remember about JS tests as well)

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -98,6 +98,112 @@ defmodule Hologram.ExJsConsistency.Erlang.FilenameTest do
...
+    end
+
+    test "empty list input" do
+      assert :filename.dirname([]) == ~c"."
+    end
+
+    test "multiple trailing slashes" do
+      assert :filename.dirname("foo/bar///") == "foo/bar"
+    end
+
```
bartblast, line 381: Please add a test for consecutive slashes in the middle of a path, e.g. `"foo//bar"`. This is an important edge case because `computeDirname` uses `lastIndexOf("/")` which would return the index of the second slash, producing `"foo/"` as the result. We need to verify this matches Erlang's actual behavior via the consistency test. Suggested test name: `"consecutive slashes in middle of path"`.

```elixir
test "consecutive slashes in middle of path" do
  assert :filename.dirname("foo//bar") == "foo/"
end
```

```javascript
it("consecutive slashes in middle of path", () => {
  const filename = Type.bitstring("foo//bar");
  const result = dirname(filename);

  assertBoxedStrictEqual(result, expected);
});
```

test/elixir/hologram/ex_js_consistency/erlang/filename_test.exs
```diff
@@ -98,6 +98,112 @@ defmodule Hologram.ExJsConsistency.Erlang.FilenameTest do
...
+    test "absolute path with trailing slash" do
+      assert :filename.dirname("/foo/") == "/foo"
+    end
+
+    test "relative path with trailing slash" do
+      assert :filename.dirname("foo/") == "foo"
+    end
+
```
bartblast, line 329: The trailing-slash tests only cover single-component paths (`"/foo/"` and `"foo/"`). Please add a test for a multi-component path with a trailing slash, e.g. `"foo/bar/baz/"`. This is important because `computeDirnameWithTrailingSlashes` returns `trimmedText` directly without further dirname processing -  we need to verify this matches Erlang for paths with more than one component. Suggested test name: `"multi-component path with trailing slash"`.

```elixir
test "multi-component path with trailing slash" do
  assert :filename.dirname("foo/bar/baz/") == "foo/bar/baz
end
```

(remember about JS test also)

assets/js/erlang/filename.mjs
```diff
@@ -158,6 +234,101 @@ const Erlang_Filename = {
...
+
+      if (lastSlashIndex === -1) return "."; // Single component - dirname is current directory
+
+      if (lastSlashIndex === 0) return "/"; // Root-relative path - dirname is root
+
+      // Multi-component path - return everything before last slash
+      return text.substring(0, lastSlashIndex);
+    };
+
+    // Handle trailing slashes: the trimmed directory path IS the parent directory
```
bartblast: This comment is misleading. The trimmed path is the path itself minus trailing slashes, not a "parent directory." For example, for input `"foo/"`, the trimmed text is `"foo"` - that's the same directory, not its parent. Consider rewording to something like:

```
// Handle trailing slashes: strip them and return the resulting path.
// A trailing slash indicates the path already refers to a directory.

bartblast:
Thank you @mward-sudo! 💜

PR #545, opened 2026-01-16
Port :sets.size/1 to JS

### LINKED ISSUE #472: Port :sets.size/1 to JS

tenkiller:
Closes #472

## Summary by CodeRabbit

* **New Features**
  * Add ability to retrieve the element count of Erlang sets, validating input and correctly handling deleted entries to ensure accurate counts.

* **Tests**
  * Add comprehensive tests covering empty sets, non-empty sets counting behavior, and error handling when given invalid inputs to ensure consistent behavior.

### DIFF
--- a/assets/js/erlang/sets.mjs
+++ b/assets/js/erlang/sets.mjs
@@ -1,5 +1,6 @@
 "use strict";
 
+import Erlang from "./erlang.mjs";
 import Erlang_Lists from "./lists.mjs";
 import Erlang_Maps from "./maps.mjs";
 import HologramInterpreterError from "../errors/interpreter_error.mjs";
@@ -220,6 +221,19 @@ const Erlang_Sets = {
   // End new/1
   // Deps: [:sets._validate_opts/1]
 
+  // Start size/1
+  "size/1": (set) => {
+    if (!Type.isMap(set)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":sets.size/1", [set]),
+      );
+    }
+
+    return Erlang["map_size/1"](set);
+  },
+  // End size/1
+  // Deps: [:erlang.map_size/1]
+
   // Start to_list/1
   "to_list/1": (set) => {
     if (!Type.isMap(set)) {
@@ -231,19 +245,6 @@ const Erlang_Sets = {
     return Erlang_Maps["keys/1"](set);
   },
   // End to_list/1
-  // Deps: [:maps.keys/1]
-
-  // Start size/1
-  "size/1": (set) => {
-    if (!Type.isMap(set)) {
-      Interpreter.raiseFunctionClauseError(
-        Interpreter.buildFunctionClauseErrorMsg(":sets.size/1", [set]),
-      );
-    }
-
-    return Type.integer(Erlang_Maps["keys/1"](set).data.length);
-  },
-  // End size/1
   // Deps: [:maps.keys/1]
 };
 
--- a/lib/hologram/compiler/call_graph.ex
+++ b/lib/hologram/compiler/call_graph.ex
@@ -71,7 +71,7 @@ defmodule Hologram.Compiler.CallGraph do
     {{:sets, :is_subset, 2}, {:sets, :is_element, 2}},
     {{:sets, :is_subset, 2}, {:sets, :to_list, 1}},
     {{:sets, :new, 1}, {:sets, :_validate_opts, 1}},
-    {{:sets, :size, 1}, {:maps, :keys, 1}},
+    {{:sets, :size, 1}, {:erlang, :map_size, 1}},
     {{:sets, :to_list, 1}, {:maps, :keys, 1}},
     {{:string, :replace, 3}, {:string, :replace, 4}},
     {{:string, :replace, 4}, {:unicode, :characters_to_binary, 1}},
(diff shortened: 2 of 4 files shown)


Lucassifoni:
I'm piggybacking here, do we need to implement kernel functions like Kernel.map_size or Kernel.tuple_size in https://github.com/bartblast/hologram/blob/dev/assets/js/elixir/kernel.mjs or there is another plan or way of doing this in Hologram ?

bartblast:
@Lucassifoni No, we don't need to implement Kernel functions like `Kernel.map_size/1` or `Kernel.tuple_size/1` in `kernel.mjs`. Hologram transpiles Elixir code, and these Kernel functions delegate to upstream Erlang functions (see: https://github.com/elixir-lang/elixir/blob/d33fd8e413fe98e0e54dde7ef94fadb272b1ef67/lib/elixir/lib/kernel.ex#L1046), which we port manually to JavaScript.

In this case, `:erlang.map_size/1` is already ported, and `:erlang.tuple_size/1` has a PR in progress.

There are some Elixir functions I've ported manually, but this is either temporary (because the required Erlang functions hadn't been ported yet) or for performance reasons (e.g., some Unicode-related functions).

Lucassifoni:
@bartblast Of course 🤦‍♂️ . Sorry for this question, it seems I was totally off the mark yesterday.

assets/js/erlang/sets.mjs
```diff
@@ -198,6 +198,19 @@ const Erlang_Sets = {
   },
   // End to_list/1
   // Deps: [:maps.keys/1]
+
+  // Start size/1
+  "size/1": (set) => {
```
bartblast: Could you order the function alphabetically within the module? (same for the test groups)

assets/js/erlang/sets.mjs
```diff
@@ -198,6 +198,19 @@ const Erlang_Sets = {
...
+    if (!Type.isMap(set)) {
+      Interpreter.raiseFunctionClauseError(
+        Interpreter.buildFunctionClauseErrorMsg(":sets.size/1", [set]),
+      );
+    }
+
+    return Type.integer(Erlang_Maps["keys/1"](set).data.length);
```
bartblast: Since `:erlang.map_size/1` is already ported, you can use it here instead of extracting keys:

```javascript
return Erlang["map_size/1"](set);
```

This avoids creating an intermediate keys array.

You will also need to:
1. Add `import Erlang from "./erlang.mjs";` at the top
2. Update the `// Deps:` comment to `[:erlang.map_size/1]`
3. Change the `call_graph.ex` entry to `{{:sets, :size, 1}, {:erlang, :map_size, 1}}`

lib/hologram/compiler/call_graph.ex
```diff
@@ -59,6 +59,7 @@ defmodule Hologram.Compiler.CallGraph do
     {{:sets, :is_subset, 2}, {:sets, :to_list, 1}},
     {{:sets, :new, 1}, {:sets, :_validate_opts, 1}},
     {{:sets, :to_list, 1}, {:maps, :keys, 1}},
+    {{:sets, :size, 1}, {:maps, :keys, 1}},
```
bartblast: This entry should be placed before the `to_list` entry to maintain alphabetical ordering (s < t).
(When you adopt the `map_size` suggestion from previous comment, also update the dependency to `{:erlang, :map_size, 1}`.)

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -402,4 +402,33 @@ defmodule Hologram.ExJsConsistency.Erlang.SetsTest do
...
   end
+
+  describe "size/1" do
+    test "returns 0 if given an empty set" do
+      set = :sets.new(version: 2)
+
+      assert :sets.size(set) == 0
+    end
+
+    test "returns correct count for non-deleted entries" do
```
bartblast: This test is redundant IMO.

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -402,4 +402,33 @@ defmodule Hologram.ExJsConsistency.Erlang.SetsTest do
...
+      set = :sets.from_list([1, 2, 3], version: 2)
+
+      assert :sets.size(set) == 3
+    end
+
+    test "excludes deleted entries from count" do
```
bartblast: This test is redundant IMO.

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -402,4 +402,33 @@ defmodule Hologram.ExJsConsistency.Erlang.SetsTest do
       end
     end
   end
+
+  describe "size/1" do
+    test "returns 0 if given an empty set" do
```
bartblast: We need another test case for a non-empty set.

test/javascript/erlang/sets_test.mjs
```diff
@@ -740,4 +740,41 @@ describe("Erlang_Sets", () => {
       );
     });
   });
+
+  describe("size/1", () => {
```
bartblast: Comments for Elixir tests apply here too.

bartblast:
Looks good, thanks @tenkiller 👍

PR #570, opened 2026-01-16
Port :os.type/0 to JS

### LINKED ISSUE #473: Port :os.type/0 to JS

tenkiller:
Closes #473 

## Summary by CodeRabbit

* **New Features**
  * Client-side Erlang OS detection added, returning platform family and name (unix, web).

* **Tests**
  * Added JavaScript and Elixir consistency tests for the new OS detection.
  * Test helpers expose the OS detection module for cross-language testing.

* **Chores**
  * Linter updated to recognize a new read-only global.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/os_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/os_test.exs
@@ -10,10 +10,10 @@ defmodule Hologram.ExJsConsistency.Erlang.OsTest do
   @moduletag :consistency
 
   describe "type/0" do
-    test "returns unix family and web name" do
+    test "returns OS family and OS name" do
       {family, name} = :os.type()
-      assert family == :unix
-      assert name == :web
+      assert family in [:win32, :unix] == true
+      assert is_atom(name) == true
     end
   end
 end
--- a/test/javascript/erlang/os_test.mjs
+++ b/test/javascript/erlang/os_test.mjs
@@ -18,7 +18,7 @@ describe("Erlang_Os", () => {
   describe("type/0", () => {
     const type = Erlang_Os["type/0"];
 
-    it("returns unix family and web name", () => {
+    it("returns OS family and OS name", () => {
       const result = type();
 
       assert.deepStrictEqual(


mward-sudo:
I'm not sure that returning the user’s OS is the correct approach in a web-based environment, as transpiled elixir->js code would be using Unix-style behaviour regardless of the user’s OS. Elixir uses this value to vary its behaviour.

I'd expect the following code would be an appropriate implementation:

```
"os/0": () => {
  return Type.tuple([Type.atom("unix"), Type.atom("web")]);
}
```

Elixir doesn't match on the second atom as it's quite variable, but we should return an appropriate os family in the first, and I believe `:unix` is correct.

Don't make changes until @bartblast reviews, but I think we should probably implement it in this manner.

### Why `{:unix, :web}` is the right choice:

- Follows the Erlang spec format
- Provides safe defaults (Unix conventions for paths and line endings)
- Won't break existing code patterns
- Semantically indicates the web environment
- Deterministic and consistent across all browsers/users

@coderabbitai Do you have an opinion on this comment?

tenkiller:
@mward-sudo I tend to agree. Similar to `:re.version/0`, we can return a hard-coded value. I'm all for whichever implementation makes better sense.

mward-sudo:
I'd remove the entire implementation and simply go with:

```js
"use strict";

import Type from "../type.mjs";

// IMPORTANT!
// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).

const Erlang_Os = {
  // Start type/0
  // Returns the hardcoded value {:unix, :web} as the code will only execute in a web context
  // Elixir / Erlang will use Unix style paths in any conditionals
  "type/0": () => {
    return Type.tuple([Type.atom("unix"), Type.atom("web")]);
  },
  // End type/0
  // Deps: []
};

export default Erlang_Os;
```

Revert your changes to other files and adjust your tests as appropriate.

Returning `:unix` as the os family type means that any conditionals using this function will use Unix style paths, which is suitable for the web.

bartblast:
@mward-sudo @tenkiller Thanks for the discussion here.

I'll admit I hesitated on this initially. I suggested matching Erlang's `:os.type/0` behavior exactly in my earlier comment: https://github.com/bartblast/hologram/issues/473#issuecomment-3711305410

My thinking at the time was:
> While we currently target browsers, proper OS detection will be valuable when we expand to desktop development. For the browser environment, the implementation should detect the actual OS using the user agent or platform APIs, just as Erlang would on different systems. This keeps the API consistent and future-proof.

However, after considering your points more carefully, I think the `{:unix, :web}` approach makes more sense:

1. This function may be used by downstream Elixir functions to decide on system-specific behaviours, which could lead to unexpected behaviours if we return the actual user OS.

2. We may need to detect the underlying system type when applying some conditional logic for system bridges in desktop and web environments, but this should be done at the framework level rather than through this function.

3. I also thought about use cases when a Hologram application genuinely needs to get info about the user's system/platform - for example, web analytics apps, platform-specific UI adjustments, or download pages offering the right installer. But we'll probably need a separate Hologram module for that. This function is too basic for that purpose anyway.

So I think @mward-sudo's suggestion makes sense here - let's hardcode `{:unix, :web}` and treat the Hologram client as a unified Unix-type platform.

@tenkiller Could you simplify the implementation as suggested? Apologies for the back-and-forth.

tenkiller:
@bartblast Sounds good, and will do. I'll have this change in before the end of the day.

mward-sudo:
It would be possible to update this function to pick up the operating environment from a Holograph framework function when the appropriate functionality is required, if it turns out that Erlang / Elixir requires OS specific handling in other environments. Picking it up directly from the browser in this function would never be suitable.

We should consider implementing OS specific behaviour in advance in the areas that Erlang uses it (such as path functions) but I think we can review that once this has landed and you've decided if `:OS.type/0` should return different OS values for future use.

bartblast:
Thank you @tenkiller! 🚀

PR #573, opened 2026-01-17
Introduce template formatter

mwmiller:
This is a second pass at introducing a template formatter for HOLO sigils (and also `.holo` files.)  This version uses Inspect Algebra for better overall results.

## Release Notes

* **New Features**
  * Added a template formatter plugin for Hologram, enabling `mix format` support for `.holo` files and `~HOLO` sigils with improved inline/block layout, attribute formatting, and whitespace normalization.

* **Bug Fixes**
  * Improved formatter correctness around `{%raw}` blocks to prevent buffered text from being lost.
  * Preserves special inviolable tags (`script`/`style`) and ensures consistent whitespace handling.

* **Tests**
  * Expanded coverage for formatter behavior and DOM parsing/traversal.


mwmiller:
This is marked as a draft until more niggles are worked out.  Upon request, I wanted to get this out publicly to make status more clear (and encourage myself to get back to work on it.)

As of now, applying the formatter in the repository itself changes 17 files.  Some of these changes I find agreeable; other changes vary from "not so much" to "breaking."

I have also used some AI assistance to get the output to this point.  I know that it contains non-idiomatic Elixir in places and I will be performing my own code review.  I hope to get it into a state where I can use the tests to avoid regressions in doing so.

I have also applied this to my own Hologram projects and found its results useful.  I think the average "in the wild" Hologram project will have less edges to navigate, so getting the output right here will be a good enough win for mainline inclusion.

bartblast (commented):
Thanks for pushing this forward, @mwmiller!

Since this is a draft I'll focus on high-level architectural decisions and formatting approach rather than implementation details.

**Architecture**

The `Algebra` module currently does two jobs (from what I understand):
1. Building a nested tree from `Parser`'s flat token list (`parse_tokens/1`, `consume_until_*`)
2. Formatting using `Inspect.Algebra`

Worth extracting tree-building into a separate `Hologram.Template.DOMTree` module:

`Parser` →  flat tokens
`DOMTree` →  nested tree
`Formatter` →  formatted string

This makes the tree structure reusable for other purposes and keeps the formatter focused on rendering.

`Inspect.Algebra` seems like a reasonable choice - we can evaluate as we go if it gives us the flexibility we need.

**Formatting heuristics**

We should base our formatting behavior on established tools like Prettier rather than inventing our own rules. The current logic expands too aggressively:

```html
<div>a</div>
```
becomes:
```html
<div>
  a
</div>
```

Prettier distinguishes between inline elements (where whitespace affects rendering) and block elements (where it doesn't), and tries to fit content on one line when possible. It's worth studying Prettier's HTML formatting approach before finalizing the heuristics here.

bartblast:
PS: The dev branch includes CI improvements that significantly speed up the GitHub Actions workflow and improve the feedback loop. You might want to merge those changes into your branch to benefit from the faster iteration.

sodapopcan:
@mwmiller Can I be of any help with getting this PR over the line?

bartblast:
@coderabbitai pause

mwmiller:
@bartblast Sorry for the noise.  I am heading back to Botswana in a week or so and thought I might have a chance to work on this.  Did a tiny bit of prep to see where I am.

arndttouby:
Hi @mwmiller — I've been using this branch's formatter in a real Phoenix + Hologram 0.11.1 project and hit a handful of issues. Sharing them here rather than opening a competing PR, since this is your work. Happy to send patches against `template_formatter` if that's useful.

**1. Attribute expressions ending in a closing brace lose it**, making the output unparsable:

```
in:  <button $click={params: %{a: 1}}>go</button>
out: <button $click={params: %{a: 1}>go</button>
```

`expression_content/1` uses `String.trim_trailing("}")`, which strips *every* trailing brace; `tighten_expression/1` then re-adds exactly one. Not length-dependent.

**2. Whitespace between inline siblings is dropped.** `<a>Home</a>\n<a>About</a>` becomes `<a>Home</a><a>About</a>` — in HTML that changes rendering. Same between an inline element and adjacent text (`<slot />\ntext` → `<slot />text`).

**3. `FunctionClauseError` on input the parser accepts:** dynamic tag names (`<{@module} a="1" />`), spread attributes (`<Widget ...{@props} />`), and unbalanced markup (`<span>a{1 + 1}c</div>`) all crash `Inspect.Algebra.string/1` or `do_consume_until_end_tag/4`.

**4. `{%raw}` inside `<script>`/`<style>` grows on every run** — the content gains two spaces of indentation per pass and never reaches a fixpoint.

To check these I built a harness that extracts every `.holo` file and `~HOLO` sigil body from this repository (459 in total) and asserts that `parse(format(x))` yields the same tree as `parse(x)`, with whitespace normalised only where HTML collapses it and expressions compared by Elixir AST. With the fixes above plus layout rules that keep single-line elements on one line, 458/459 pass and 391 are left byte-identical. The remaining one is the intentionally unbalanced parser fixture.

Would you like the patches, the harness, or both?

arndttouby:
@bartblast — a direction question for you, since it decides where effort is worth spending: is a `mix format` plugin working on the parser's token stream the route you want for template formatting, or does the `DOMTree` work in #581 supersede this approach? The defects above are all fixable on the current basis, but several of them come from reconstructing structure from a flat token list, so a tree-based printer would make them disappear rather than need patching. Happy to contribute either way.

PR #581, opened 2026-01-18
Introduce DOMTree for hierarchical template representation

mwmiller:
This module transforms the flat output of the template parser into a hierarchical, indexed tree structure, enabling advanced template transformations and detailed developer feedback.

Key features for consumers:
- **Map/Reduce Traversal:** The traverse/3 function provides a powerful way to transform the tree and aggregate state simultaneously using a guarded, two-arity callback.
- **Precise Debugging:** Each node implements the String.Chars protocol for human-readable logging (e.g. "<div (CID: 0.1)>" or "{%if true}") and find_by_cid/2 for targeted node extraction.
- **Descriptive Error Messages:** Error reporting pinpoints the exact problematic node and its position in the hierarchy using CIDs, reducing ambiguity in complex templates.
- **Flexible Attribute Access:** get_attribute/3 retrieves values as flattened strings by default, or as raw metadata parts when needed.
- **Stable Node Identity:** Automatically generates hierarchical Component IDs (CIDs) while respecting and promoting explicit developer-provided 'cid' attributes in markup.
- **Ergonomic Helpers:** A full suite of predicates (element?/1, block?/1, component?/1, etc.) simplifies pattern matching and tree consumption.

The implementation handles multiple HTML fragments, features robust non-fatal error accumulation, and is backed by a comprehensive test suite covering the full lifecycle from parse to inspection.

## Summary by CodeRabbit

* **New Features**
  * Added a public DOM tree API for parsed templates enabling construction, traversal, and inspection of DOM-like nodes with metadata.

* **Bug Fixes**
  * Improved detection and clearer errors for unclosed, mismatched, or unexpected tags (including else cases).

* **Documentation**
  * Added public docs and type specs describing the DOM API and traversal behavior.

* **Tests**
  * Comprehensive tests covering parsing scenarios, attributes, blocks, traversal, edge cases, and error handling.

### DIFF
--- a/lib/hologram/template/dom_tree.ex
+++ b/lib/hologram/template/dom_tree.ex
@@ -1,0 +1,216 @@
+defmodule Hologram.Template.DOMTree do
+  @moduledoc """
+  Converts the output of the template parser into a DOMTree structure.
+  """
+
+  alias Hologram.Template.Helpers
+  alias Hologram.Template.Parser
+
+  defmodule Node do
+    @type t :: %__MODULE__{
+            type: :component | :element | :text | :expression | :block | :doctype | :comment,
+            name: String.t() | nil,
+            attributes: list({String.t(), any()}) | nil,
+            children: Hologram.Template.DOMTree.nodes() | nil,
+            content: String.t() | any() | nil,
+            module: module() | nil
+          }
+
+    defstruct [:type, :name, :content, :module, attributes: [], children: []]
+  end
+
+  @type nodes :: list(Node.t())
+
+  @doc """
+  Converts a list of parsed tags into a list of DOMTree nodes.
+  """
+  @spec from_parse(list(Parser.parsed_tag())) :: {:ok, nodes()} | {:error, any()}
+  def from_parse(tags) do
+    process(tags, [], [])
+  end
+
+  # End of input
+  defp process([], [], roots) do
+    {:ok, Enum.reverse(roots)}
+  end
+
+  defp process([], [{node, _children} | tail], _roots) do
+    {:error, {:unclosed_tag, node, parents(tail)}}
+  end
+
+  # Text
+  defp process([{:text, content} | rest], stack, roots) do
+    node = %Node{type: :text, content: content}
+    add_node(node, rest, stack, roots)
+  end
+
+  # Start Tag
+  defp process([{:start_tag, {tag_name, attrs}} | rest], stack, roots) do
+    node = %Node{
+      type: Helpers.tag_type(tag_name),
+      name: tag_name,
+      attributes: attrs
+    }
+
+    process(rest, [{node, []} | stack], roots)
+  end
+
+  # Self Closing Tag
+  defp process([{:self_closing_tag, {tag_name, attrs}} | rest], stack, roots) do
+    node = %Node{
+      type: Helpers.tag_type(tag_name),
+      name: tag_name,
+      attributes: attrs
+    }
+
+    add_node(node, rest, stack, roots)
+  end
+
+  # End Tag
+  defp process([{:end_tag, tag_name} | rest], stack, roots) do
+    handle_end_tag(tag_name, :html, rest, stack, roots)
+  end
+
+  # Block Start: Else
+  defp process([{:block_start, "else"} | _rest], [], _roots) do
+    {:error, {:unexpected_tag, "else", []}}
+  end
+
+  defp process([{:block_start, "else"} | rest], stack, roots) do
+    case stack do
+      [{%Node{type: :block, name: "if"}, _children} | _tail] ->
+        node = %Node{type: :block, name: "else"}
+        process(rest, [{node, []} | stack], roots)
+
+      _stack ->
+        {:error, {:unexpected_tag, "else", parents(stack)}}
+    end
+  end
+
+  # Block Start: Raw
+  defp process([{:block_start, "raw"} | rest], stack, roots) do
+    node = %Node{type: :block, name: "raw"}
+    process(rest, [{node, []} | stack], roots)
+  end
+
+  # Block Start: Generic
+  defp process([{:block_start, {name, expr}} | rest], stack, roots) do
+    node = %Node{
+      type: :block,
+      name: name,
+      content: expr
+    }
+
+    process(rest, [{node, []} | stack], roots)
+  end
+
+  # Block End
+  defp process([{:block_end, name} | rest], stack, roots) do
+    handle_end_tag(name, :block, rest, stack, roots)
+  end
+
+  # Expression
+  defp process([{:expression, content} | rest], stack, roots) do
+    node = %Node{
+      type: :expression,
+      content: content
+    }
+
+    add_node(node, rest, stack, roots)
+  end
+
+  # Doctype
+  defp process([{:doctype, content} | rest], stack, roots) do
+    node = %Node{type: :doctype, content: content}
+    add_node(node, rest, stack, roots)
+  end
+
+  # Comment Start
+  defp process([:public_comment_start | rest], stack, roots) do
+    node = %Node{type: :comment}
+    process(rest, [{node, []} | stack], roots)
+  end
+
+  # Comment End
+  defp process([:public_comment_end | rest], stack, roots) do
+    case stack do
+      [{%{type: :comment} = node, children} | stack_tail] ->
+        finished_node = %{node | children: Enum.reverse(children)}
+        add_node(finished_node, rest, stack_tail, roots)
+
+      _stack ->
+        {:error, {:unexpected_closing_tag, "-->", :html, parents(stack)}}
+    end
+  end
+
+  # Helpers
+
+  defp add_node(node, rest, [], roots) do
+    process(rest, [], [node | roots])
+  end
+
+  defp add_node(node, rest, [{parent, children} | tail], roots) do
+    new_stack = [{parent, [node | children]} | tail]
+    process(rest, new_stack, roots)
+  end
+
+  defp handle_end_tag(tag_name, kind, rest, stack, roots) do
+    case stack do
+      [{%{name: ^tag_name} = node, children} | stack_tail] ->
+        finished_node = %{node | children: Enum.reverse(children)}
+        add_node(finished_node, rest, stack_tail, roots)
+
+      [{%{name: "else"}, _else_children} | [{%{name: "if"}, _if_children} | _if_tail]]
+      when tag_name == "if" ->
+        handle_implicit_else_closing(rest, stack, roots)
+
+      _stack ->
+        {:error, {:unexpected_closing_tag, tag_name, kind, parents(stack)}}
+    end
+  end
+
+  defp handle_implicit_else_closing(rest, stack, roots) do
+    [
+      {%{name: "else"} = else_node, else_children}
+      | [{%{name: "if"} = if_node, if_children} | if_tail]
+    ] = stack
+
+    finished_else = %{else_node | children: Enum.reverse(else_children)}
+    new_if_children = [finished_else | if_children]
+
+    finished_if = %{if_node | children: Enum.reverse(new_if_children)}
+    add_node(finished_if, rest, if_tail, roots)
+  end
+
+  defp parents(stack) do
+    Enum.map(stack, fn {node, _children} -> node end)
+  end
+
+  @doc """
+  Traverses the DOM tree(s) in a map/reduce style (depth-first, pre-order).
+
+  The callback function should take a node and an accumulator, and return
+  a tuple `{new_node, new_accumulator}`. This is useful for transforming the
+  tree while collecting data, such as a list of used components.
+  """
+  @spec traverse(Node.t() | nodes(), any(), (Node.t(), any() -> {Node.t(), any()})) ::
+          {Node.t() | nodes(), any()}
+  def traverse(nodes, acc, callback) when is_list(nodes) and is_function(callback, 2) do
+    Enum.map_reduce(nodes, acc, fn node, current_acc ->
+      traverse(node, current_acc, callback)
+    end)
+  end
+
+  def traverse(node, acc, callback) when is_function(callback, 2) do
+    {new_node, new_acc} = callback.(node, acc)
+
+    case new_node.children do
+      children when is_list(children) ->
+        {new_children, final_acc} = traverse(children, new_acc, callback)
+        {%{new_node | children: new_children}, final_acc}
+
+      _children ->
+        {new_node, new_acc}
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


mwmiller:
This PR acts a precursor to #573.  It allows me to focus on each bit of the transformation in isolation which is better for my sanity.  If we are happy with this change, any eventual formatter can use this to restyle/regenerate the semantically important bits of the source to create formatter output.

Some of the bits around attempting to maintain an "understanding of user intent" might become more difficult this way, while others around opinionated styling might become easier.

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,513 @@
...
+    "<!-- comment --> (CID: #{cid})"
+  end
+
+  @spec truncate(String.t()) :: String.t()
+  defp truncate(content) do
+    case String.length(content) > 20 do
+      true -> String.slice(content, 0..17) <> "..."
+      false -> content
+    end
+  end
```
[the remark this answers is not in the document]
mwmiller: Folded adjustments into commit.

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,507 @@
...
+
+  defp explain_error({:unclosed_tag, node, context}) do
+    "Unclosed #{node}" <> format_context(context)
+  end
+
+  defp explain_error({:unexpected_closing_tag, tag, context}) do
+    tag_str = if tag == "-->", do: "-->", else: "</#{tag}>"
+    "Unexpected closing tag #{tag_str}" <> format_context(context)
+  end
```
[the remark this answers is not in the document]
mwmiller: Folded in as others.

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
+defmodule Hologram.Template.DOMTree do
+  @moduledoc """
+  Converts the output of the template parser into a DOMTree structure.
+  """
+
+  alias Hologram.Template.Helpers
+  alias Hologram.Template.Parser
+
+  @type t :: %__MODULE__{
```
bartblast: This should probably be called `Hologram.Template.DOMTree.Node`, what do you think?

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
...
+          type: :component | :element | :text | :expression | :block | :doctype | :comment,
+          name: String.t() | nil,
+          attributes: list({String.t(), any()}) | nil,
+          children: list(t()) | nil,
+          content: String.t() | any() | nil,
+          module: module() | nil,
+          cid: String.t() | nil
```
bartblast: Could you explain the need for assigning CIDs to every node type? In Hologram, CID stands for "Component ID" and is used only for components, not for regular HTML elements, text nodes, etc. This naming could cause confusion.

Also, error messages like "Unclosed <span (CID: 1.0.2)>, found within section > div (CID: 1.0)." aren't meaningful to end users. Compare with the parser's contextual approach:
```
Close the start tag with '>' character.
<div
    ^
```

Could you confirm that we'd be able to achieve similar error reporting if we pass position metadata through parser tags? But let's not tackle that in this PR - for the formatter project, let's focus on formatting functionality and keep error messages basic for now.

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
...
+  alias Hologram.Template.Helpers
+  alias Hologram.Template.Parser
+
+  @type t :: %__MODULE__{
+          type: :component | :element | :text | :expression | :block | :doctype | :comment,
+          name: String.t() | nil,
+          attributes: list({String.t(), any()}) | nil,
+          children: list(t()) | nil,
+          content: String.t() | any() | nil,
+          module: module() | nil,
```
bartblast: What is the `module` field intended for? I don't see it being populated anywhere in the code.

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
...
+
+    @type t :: %__MODULE__{
+            roots: list(Hologram.Template.DOMTree.t()),
+            root_count: non_neg_integer(),
+            stack:
+              list(
+                {Hologram.Template.DOMTree.t(), list(Hologram.Template.DOMTree.t()),
+                 non_neg_integer()}
+              ),
+            errors: list(any())
```
bartblast: What do you think about using fail-fast instead of error accumulation here?

A few considerations:
- Consistency - The parser uses fail-fast (raises `TemplateSyntaxError`)
- Cascading errors - Structural errors tend to cascade. E.g., with `<div><span></div>`, the unclosed `<span>` might cause multiple downstream errors that are really just consequences of the first mistake
- Simpler implementation - No need for the errors list, reversing, and end-of-stream checking
- Context - For a formatter, if the template has structural issues, we probably shouldn't format it at all - just show the first problem

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
...
+          }
+  end
+
+  @doc """
+  Creates a new DOMTree struct from a keyword list of attributes.
+  """
+  @spec new(keyword()) :: t()
+  def new(attrs \\ []) do
```
bartblast: What use case will this function serve in the formatting or template validation process?

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
...
+  @doc """
+  Traverses the DOM tree(s) in a map/reduce style (depth-first, pre-order).
+
+  The callback function should take a node and an accumulator, and return
+  a tuple `{new_node, new_accumulator}`. This is useful for transforming the
+  tree while collecting data, such as a list of used components.
+  """
+  @spec traverse(t() | list(t()), any(), (t(), any() -> {t(), any()})) ::
+          {t() | list(t()), any()}
+  def traverse(nodes, acc, callback) when is_list(nodes) and is_function(callback, 2) do
```
bartblast, line 198: Will all these public functions (`traverse/3,` type predicates, etc.) actually be needed for the formatter, or were they created speculatively? Could we start with just `from_parse/1` and add others when there's a clear need?

lib/hologram/template/dom_tree.ex
```diff
@@ -0,0 +1,539 @@
...
+  def find_by_cid(%__MODULE__{cid: target_cid} = node, target_cid), do: node
+
+  def find_by_cid(%__MODULE__{children: children}, target_cid) when is_list(children) do
+    find_by_cid(children, target_cid)
+  end
+
+  def find_by_cid(_node, _target_cid), do: nil
+end
+
+defimpl String.Chars, for: Hologram.Template.DOMTree do
```
bartblast: If this is needed (though it may not be, depending on the CID discussion above), it should go in a separate file with separate tests IMO.

test/elixir/hologram/template/dom_tree_test.exs
```diff
@@ -0,0 +1,628 @@
+defmodule Hologram.Template.DOMTreeTest do
+  use ExUnit.Case, async: true
+  alias Hologram.Template.DOMTree
+  alias Hologram.Template.Parser
+
+  test "parses simple text" do
```
bartblast, line 9: Suggestion for test style - in my opinion it would be cleaner to:

  1. Use the `parsed_tags/1` test helper instead of calling `Parser.parse_markup(markup)` each time
  2. Import `DOMTree.from_parse/1` so you can call it directly (that's the pattern used in other tests)
  3. Compare using equality instead of pattern matching:

```elixir
test "parses simple text" do
  result =
    "Hello world"
    |> parsed_tags()
    |> from_parse()

  assert result == {:ok, [%DOMTree{type: :text, content: "Hello world"}]}
end
```

test/elixir/hologram/template/dom_tree_test.exs
```diff
@@ -0,0 +1,628 @@
...
+    [span] = div.children
+    assert span.cid == "nested_id"
+    assert span.attributes == []
+  end
+
+  test "parses expression" do
+    markup = "{1 + 1}"
+    tags = Parser.parse_markup(markup)
+
+    assert {:ok, [%DOMTree{type: :expression, content: "1 + 1"}]} = DOMTree.from_parse(tags)
```
bartblast: Consider the `:content` field - should we be stripping the braces here? The {...} syntax is used in multiple contexts - template interpolation, component props, actions, etc. The parser already preserves the full expression as `{1 + 1}` - stripping the braces seems to lose information that might be needed for formatting.

test/elixir/hologram/template/dom_tree_test.exs
```diff
@@ -0,0 +1,628 @@
...
+  test "parses if block" do
+    markup = "{%if true}content{/if}"
+    tags = Parser.parse_markup(markup)
+
+    assert {:ok,
+            [
+              %DOMTree{
+                type: :block,
+                name: "if",
+                content: "true",
```
bartblast: Here the expression is correctly treated as `true` (in reference to the previous comment about stripping the braces).

test/elixir/hologram/template/dom_tree_test.exs
```diff
@@ -0,0 +1,628 @@
...
+            </li>
+          {/for}
+        </ul>
+      {%else}
+        <p>Please <a href="/login">login</a>.</p>
+      {/if}
+      <footer>Page {1 + 2}</footer>
+    </div>
+    """
+
```
bartblast: For complex tests like this, wouldn't it be simpler to compare the full result directly rather than picking apart the tree with many intermediate assertions? The current approach filters out text nodes and is hard to follow. A direct comparison would be more complete and easier to maintain.

bartblast (changes requested):
Thanks! Some foundational questions to resolve first.

lib/hologram/template/dom_tree.ex
mwmiller: I admit I got a bit focused on error location here at the loss of the ability to reproduce the source more exactly.

lib/hologram/template/dom_tree.ex
mwmiller: I believed at the start that I could have namespacing issues. I never cleaned this up.

lib/hologram/template/dom_tree.ex
mwmiller: I'm not generally a huge fan of fail-fast, because I believe it can be misleading about the underlying issues.  On the other hand, errors in parsing/handling HTML will tend to happen top-down and it's the repository standard, so sure.

lib/hologram/template/dom_tree.ex
mwmiller: I was trying to improve on the (frankly poor) error reporting from the parser.  I didn't achieve all that I hoped in large part because information disappears at each level.  I'll probably give up the dream.

lib/hologram/template/dom_tree.ex
mwmiller: I would like to make sure it is clear that I have a working formatter which does not require this DOM parsing.  I am doing the DOM parsing on request to make it usable for things "beyond formatting."  I don't mind helping out (and it might even improve the straight-line readability or nesting logic of the formatter), but I am responding to a request for an intermediate representation which is, somehow, useful.  I was trying to fulfill that brief.

bartblast:
@mwmiller Thanks for explaining your thinking on these points, and I want to make sure I'm being clear: I appreciate the effort you've put into this. The questions I raised aren't meant as criticism of the work itself - they're about aligning on the foundation before we build on top of it.

**Fail-fast vs error accumulation:** My preference here isn't about codebase consistency - it's about how structural errors cascade. With something like `<div><span></div>`, the unclosed `<span>` would likely produce multiple downstream errors that are just noise from the first mistake. This is the standard approach for formatters - Prettier, rustfmt, and gofmt all refuse to format syntactically invalid code rather than attempting recovery. It's also how the Hologram template parser already works (raises `TemplateSyntaxError` immediately). Formatting is fast - just fix the first error and re-run.

**The error reporting:** "Give up the dream" made me sad to read ;) - please don't! The current Hologram error reporting is genuinely bad DX and we need to fix it. But let's tackle that as a separate effort where we can focus on it properly. For this project, let's just focus on pushing the formatter across the finish line.

Looking back, maybe I focused too much on reviewing every detail of this module. My main concern was making sure the DOM tree approach we land on will be extendable for better error reporting later, and that it makes the formatter implementation itself easier. If it does both, we're in good shape.

What would help me most: could you outline which parts of this PR are essential for the formatter to work, vs. which were exploratory/future-facing? Or do you suggest we just merge it as-is for now and revisit as we work on improving template errors later?

Let me know what you think :)

mwmiller:
I think I can greatly simplify this PR.

I also want to apologize up front.  I think I sounded a bit peevish.  That is just leaking through from real life stressors.  It's not a function of this Hologram work and it shouldn't be appear in my comments.

In any case, I am going to work on an improved version which scales back the accessories and focuses on use cases I can justify to myself up front.

When/if it is incorporated we can worry about how it might be helpful in other cases.

As is my wont, I will likely do this in a rewritten commit.  The thought/iteration process here doesn't seem likely to be helpful to mainline history.

mwmiller:
@bartblast I have ripped out most of what is unneeded for formatting purposes at this time.

I'm continuing to work on this PR.  I am applying changes recommended by the AI and yourself.  I expect it to be ready for review/consideration sometime tomorrow.

When/if it is merged I will rebase the formatter work on top of this and try to bring it more into an opinionated user-intent preserving style.  As I have mentioned this can have some issues with:

- Lost information from the toolchain adjustments
- Deterministic output with semantically identical sources with different inferred user intent

I hope to make it palatable quickly.  I don't especially want to rush it to mainline, because a formatter which makes wide but minor adjustments ex-post-facto is even more annoying than not having a formatter available.

bartblast:
@mwmiller

> I also want to apologize up front. I think I sounded a bit peevish. That is just leaking through from real life stressors. It's not a function of this Hologram work and it shouldn't be appear in my comments.

No worries at all! I honestly didn't read it as peevish - maybe a bit disappointed or resigned, but that's the eternal mystery of text-based communication :) You never quite know what the person on the other side of the screen is actually feeling. For context, I've had genuinely rough exchanges in reviews on this project. Yours didn't register on that scale at all. And I should apologize for being too dry or nitpicky sometimes myself - when you're juggling a lot of review threads, it's easy to forget there's a human putting real effort into every contribution.

> I don't especially want to rush it to mainline, because a formatter which makes wide but minor adjustments ex-post-facto is even more annoying than not having a formatter available.

Agreed on not rushing this. We can easily push it to the next release if you prefer. And don't worry about holding things up - I'm shortening release cycles, so minor releases will usually go out every 1-2 weeks depending on what features are ready.

Just ping me when it's ready for another look!

bartblast:
Hey Matt, just checking in - how's it going with this PR? Is there anything you need from me, any questions or additional context I can help with?

bartblast:
@coderabbitai pause

PR #582, opened 2026-01-18
Add `sa` mix alias for static analysis

mwmiller:
A simplified way to run all of the included static analysis tasks for this repository.  It ought to make it more ergonomic for contributors to avoid back and forth with the CI checks.

## Summary by CodeRabbit

* **Chores**
  * Added a new development alias group for running multiple quality-check and analysis tasks including code formatting checks, linting, type checking, code analysis, and dependency auditing.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -16,6 +16,19 @@ defmodule Hologram.MixProject do
         "cmd assets/node_modules/.bin/prettier '*.yml' '.github/**' 'assets/*.json' 'assets/*.mjs' 'assets/js/**' 'benchmarks/javascript/**' 'scripts/**' 'test/javascript/**' --config 'assets/.prettierrc.json' --write",
       "format.js.check":
         "cmd assets/node_modules/.bin/prettier '*.yml' '.github/**' 'assets/*.json' 'assets/*.mjs' 'assets/js/**' 'benchmarks/javascript/**' 'scripts/**' 'test/javascript/**' --check --config 'assets/.prettierrc.json' --no-error-on-unmatched-pattern",
+      sa: [
+        "format --check-formatted",
+        "format.js.check",
+        "cmd cd test/features && mix format --check-formatted",
+        "credo",
+        "dialyzer",
+        "doctor",
+        "sobelow --config",
+        "deps.audit",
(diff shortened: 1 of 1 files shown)


bartblast:
Thanks for the suggestion! Are you aware the project already has `mix check` configured (via ex_check)? It runs all these checks and more. See `.check.exs` for the full list.

mwmiller:
That is not quite what I was looking for but it already exists which makes it better!

bartblast:
@mwmiller What were you originally looking for? If there's a gap in the current tooling, I'd be happy to hear about it.

mwmiller:
I wanted three things with the `sa` alias:

- Something which was (as much as possible) free of side-effects.  I run it.. it passes or fails, but either way "the world" is unchanged
- Something covers the majority of the surface area of the CI. If I run and it passes I'm highly unlikely to run into "style" issues rejections from the GH Actions.
- Something which runs quickly and effectively on my machine.  The full test suite does not pass on my machine out of the box and it takes a long time doing NPM stuff which seems like duplicated work in order to eventually fail in an expected way[0].

I guess the major issue for me is that an early fail and bail in the test suite won't let me continue on to potential unexpected failures in other areas.

[0] The below is that expected way.  I never have considered it important or interesting enough to look into why this suddenly started failing a couple months back:

 5) test install_js_deps/1 raises RuntimeError if npm install command fails (Hologram.CompilerTest)
     test/elixir/hologram/compiler_test.exs:638
     Expected exception RuntimeError but got File.Error (could not read file "/Users/matt/git/hologram/tmp/tests/compiler/install_js_deps_1/assets/package.json": no such file or directory)
     code: assert_raise RuntimeError, "npm install command failed", fn ->
     stacktrace:
       (elixir 1.18.2) lib/file.ex:385: File.read!/1
       (hologram 0.6.6) lib/hologram/compiler.ex:605: Hologram.Compiler.get_package_json_digest/1
       (hologram 0.6.6) lib/hologram/compiler.ex:422: Hologram.Compiler.install_js_deps/2
       test/elixir/hologram/compiler_test.exs:645: (test)

PR #598, opened 2026-01-20
Port :filelib.safe_relative_path/2 to JS

### LINKED ISSUE #597: Port :filelib.safe_relative_path/2 to JS
```
-spec safe_relative_path(Filename, Cwd) -> unsafe | SafeFilename
                            when
                                Filename :: filename_all(),
                                Cwd :: filename_all(),
                                SafeFilename :: filename_all().
```

- [Erlang docs](https://www.erlang.org/doc/apps/stdlib/filelib.html#safe_relative_path/2)

mward-sudo:
Closes #597 

## Summary by CodeRabbit

* **New Features**
  * Added a safe path-sanitization utility that normalizes relative paths and prevents unsafe traversals or absolute-path results.

* **Tests**
  * Added comprehensive JavaScript and Elixir test suites covering many input types, edge cases, type-preserving behavior, and error conditions.
  * Updated test helpers to expose the new utility to the test environment.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/filelib_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/filelib_test.exs
@@ -42,6 +42,22 @@ defmodule Hologram.ExJsConsistency.Erlang.FilelibTest do
       assert :filelib.safe_relative_path("", "/home/user") == []
     end
 
+    test "just ." do
+      assert :filelib.safe_relative_path(".", "/home") == []
+    end
+
+    test "just .." do
+      assert :filelib.safe_relative_path("..", "/home") == :unsafe
+    end
+
+    test ".. in the middle resolving to sibling" do
+      assert :filelib.safe_relative_path("a/../b", "/home") == "b"
+    end
+
+    test "atom input" do
+      assert :filelib.safe_relative_path(:dir, "/home") == ~c"dir"
+    end
+
     test "charlist relative path" do
       assert :filelib.safe_relative_path(~c"dir", ~c"/") == ~c"dir"
     end
--- a/test/javascript/erlang/filelib_test.mjs
+++ b/test/javascript/erlang/filelib_test.mjs
@@ -92,6 +92,42 @@ describe("Erlang_Filelib", () => {
       assert.deepStrictEqual(result, expected);
     });
 
+    it("just .", () => {
+      const filename = Type.bitstring(".");
+      const cwd = Type.bitstring("/home/user");
+      const result = safeRelativePath(filename, cwd);
+      const expected = Type.list([]);
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it("just ..", () => {
+      const filename = Type.bitstring("..");
+      const cwd = Type.bitstring("/home/user");
+      const result = safeRelativePath(filename, cwd);
+      const expected = Type.atom("unsafe");
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it(".. in the middle resolving to sibling", () => {
+      const filename = Type.bitstring("a/../b");
+      const cwd = Type.bitstring("/home/user");
+      const result = safeRelativePath(filename, cwd);
+      const expected = Type.bitstring("b");
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it("atom input", () => {
+      const filename = Type.atom("dir");
+      const cwd = Type.bitstring("/home/user");
+      const result = safeRelativePath(filename, cwd);
+      const expected = Type.charlist("dir");
+
+      assert.deepStrictEqual(result, expected);
+    });
+
     it("charlist relative path", () => {
       // "dir"
       const filename = Type.list([


assets/js/erlang/filelib.mjs
```diff
@@ -0,0 +1,136 @@
...
+      const withPart = extendArray(bytes, part);
+      const isLastPart = index === sanitizedParts.parts.length - 1;
+      return isLastPart ? withPart : [...withPart, DIR_SEPARATOR_BYTE];
+    }, []);
+
+    // Build result based on input type
+    return isBinaryInput
+      ? buildBinaryResult(joinedBytes)
+      : buildCharlistResult(joinedBytes);
+  },
```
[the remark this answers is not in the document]
mward-sudo: # Refutation: CodeRabbitAI's Claim About Cwd Parameter Usage

## Summary

CodeRabbitAI claims that `safe_relative_path/2` should convert absolute paths that fall within `Cwd` to relative paths. This is **incorrect**. Erlang's `:filelib.safe_relative_path/2` rejects all absolute paths unconditionally and does not use the `Cwd` parameter for validation.

## CodeRabbitAI's Claim

> "The _cwd parameter must be used to check if absolute Filename is within Cwd and convert it to a relative path.
>
> According to Erlang/OTP filelib documentation, Cwd is the base directory against which the relative path is resolved. The current implementation ignores _cwd entirely and rejects any absolute path unconditionally. Instead, if Filename is absolute and falls within the Cwd directory, it should be converted to a relative path; only truly unsafe paths (outside Cwd or traversing parent directories) should return "unsafe". For example, safe_relative_path("/home/user/dir", "/home/user") should return a relative path, not "unsafe"."

## Actual Erlang Behavior (Tested 2026-01-20)

### Absolute Paths Are Always Unsafe

Regardless of `Cwd` value:

```elixir
:filelib.safe_relative_path("/home/user/dir", "/home/user") # → :unsafe
:filelib.safe_relative_path("/a/b", "/a") # → :unsafe
:filelib.safe_relative_path("/abs/path", "/home/user") # → :unsafe
```

**Result:** All absolute paths return `:unsafe`, contradicting CodeRabbitAI's expected behavior.

### Cwd Parameter Is Ignored

Testing with various `Cwd` formats all return identical results for the same `Filename`:

```elixir
# Relative path "dir" tested with different Cwd values:
:filelib.safe_relative_path("dir", "/home/user") # → ~c"dir"
:filelib.safe_relative_path("dir", "relative") # → ~c"dir"
:filelib.safe_relative_path("dir", ".") # → ~c"dir"
:filelib.safe_relative_path("dir", "..") # → ~c"dir"
:filelib.safe_relative_path("dir", "") # → ~c"dir"
:filelib.safe_relative_path("dir", []) # → ~c"dir"
```

**Result:** Output is identical regardless of `Cwd` type or value. If `Cwd` were used for validation, results would differ. They don't.

## What Erlang Actually Does

1. **Accepts relative paths** — Returns the normalized relative path or `:unsafe` if attempting to escape via `..`
2. **Rejects all absolute paths** — Returns `:unsafe` unconditionally, regardless of `Cwd` value
3. **Cwd parameter is unused** — Present for API compatibility only; has zero effect on the result

## Verification: Our Implementation Matches Erlang

### Test Results

- **JavaScript tests:** 2982 passing (including 20 filelib-specific tests)
- **Elixir consistency tests:** 4248 tests, 0 failures (20 of which directly call `:filelib.safe_relative_path/2`)

### Example Test Parity

Erlang behavior vs. our JS port on absolute paths:

```
Erlang: :filelib.safe_relative_path("/a/b", "/a") → :unsafe
JS:     Erlang_Filelib["safe_relative_path/2"](Type.bitstring("/a/b"), Type.bitstring("/a")) → Type.atom("unsafe")
```

All tests pass because our implementation correctly mirrors Erlang's behavior.

## Conclusion

CodeRabbitAI is describing a **different function** than Erlang's `:filelib.safe_relative_path/2`. The implementation is correct as-is. The `Cwd` parameter exists but is intentionally unused, as verified by direct testing against Erlang/OTP.

The current implementation should not be changed to match CodeRabbitAI's interpretation, as doing so would break compatibility with actual Erlang behavior.

assets/js/erlang/filelib.mjs
```diff
@@ -0,0 +1,136 @@
+"use strict";
+
+import Bitstring from "../bitstring.mjs";
+import Type from "../type.mjs";
+import Erlang_Filename from "./filename.mjs";
```
bartblast, line 4: Could you order imports alphabetically?

assets/js/erlang/filelib.mjs
```diff
@@ -0,0 +1,136 @@
...
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const Erlang_Filelib = {
```
bartblast, line 12: `Erlang_Filelib` needs to be imported and registered in `helpers.mjs`, following the same pattern as every other Erlang module. Could you add:

```js
import Erlang_Filelib from "../../../assets/js/erlang/filelib.mjs";
```

in the imports (alphabetically), and:

```js
globalThis.Erlang_Filelib = Erlang_Filelib;
```

in `defineGlobalErlangAndElixirModules()` (alphabetically).

assets/js/erlang/filelib.mjs
```diff
@@ -0,0 +1,136 @@
...
+  // Start safe_relative_path/2
+  "safe_relative_path/2": (filename, _cwd) => {
```
bartblast: Two issues with the `cwd` parameter:

1. **No type validation:** `cwd` is never validated. In real Erlang, invalid `cwd` types (integers, tuples, etc.) would eventually surface as errors during symlink resolution via `file:read_link/1`. Here, passing `123` or `null` as `cwd` is silently accepted. Please add basic type validation to match OTP's accepted types (`filename_all()` - binary, charlist, or atom), even if the value itself isn't used.

2. **No documentation of why it's unused:** In the OTP implementation, `cwd` is joined with the accumulated path segments to construct absolute filesystem paths, which are then passed to `file:read_link/1` to detect symlinks that could escape the relative root (see `srp_segment/4` in [OTP filelib.erl](https://github.com/erlang/otp/blob/master/lib/stdlib/src/filelib.erl)). Since there's no filesystem access in a browser JS environment, this can't be replicated - but please add a comment explaining the limitation so future maintainers understand it's intentional:

```js
// Note: In OTP, cwd is used to build absolute paths for each segment in order
// to call :file.read_link/1 and detect symlinks that would escape the relative
// root (see srp_segment/4 in filelib.erl). This is not applicable in a browser
// environment, so only pure path sanitization (removing ".", resolving "..",
// rejecting absolute paths) is performed here.
```

test/javascript/erlang/filelib_test.mjs
```diff
@@ -0,0 +1,239 @@
...
+      const filename = Type.bitstring("dir");
+      const cwd = Type.bitstring("");
+      const result = safeRelativePath(filename, cwd);
+      const expected = Type.bitstring("dir");
+
+      assert.deepStrictEqual(result, expected);
+    });
+
+    it("cwd with empty list is normalized to dot", () => {
+      const filename = Type.bitstring("dir");
```
bartblast: `Type.charlist("dir")` for both `filename` and `expected` - to match the Elixir consistency test.

test/elixir/hologram/ex_js_consistency/erlang/filelib_test.exs
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    test "path with trailing slashes" do
+      assert :filelib.safe_relative_path("dir/sub/", "/home") == "dir/sub"
+    end
+
+    test "path with only . components" do
+      assert :filelib.safe_relative_path("./.././.", "/home") == :unsafe
+    end
+
```
bartblast, line 96: Please add the missing test cases (remember about JS tests also):

1. **Just `".."`** - should return `:unsafe`
2. **Just `"."`** - should return empty list `[]`
3. **`"a/../b"` (parent dir resolving to sibling)** - should return `"b"`
4. **Atom input** (e.g. `:dir`) - Erlang accepts atoms and returns a charlist. Should verify the JS implementation handles this correctly.

```elixir
test "just .." do
  assert :filelib.safe_relative_path("..", "/home") == :unsafe
end

test "just ." do
  assert :filelib.safe_relative_path(".", "/home") == []
end

test ".. in the middle resolving to sibling" do
  assert :filelib.safe_relative_path("a/../b", "/home") == "b"
end

test "atom input" do
  assert :filelib.safe_relative_path(:dir, "/home") == ~c"dir"
end

bartblast:
Looks good, thank you @mward-sudo! 🙏

PR #605, opened 2026-01-22
Port :unicode_util.gc/1 to JS

### LINKED ISSUE #604: Port :unicode_util.gc/1 to JS
Can't find any docs for this function, so will have to feel it out.

mward-sudo:
Closes #604 

## Summary by CodeRabbit

* **New Features**
  * Added a public grapheme-aware text extraction API that returns a leading grapheme cluster plus the remainder and handles binaries, lists, mixed tails, combining marks, multi-codepoint emoji (including flag sequences), and boundary edge cases.

* **Tests**
  * Added comprehensive JavaScript and Elixir test suites covering binaries, lists, combining-mark grouping, emoji scenarios, empty/single-codepoint inputs, nested/mixed structures, misalignment, and error cases.

### DIFF
--- a/test/javascript/erlang/unicode_util_test.mjs
+++ b/test/javascript/erlang/unicode_util_test.mjs
@@ -1011,8 +1011,16 @@ describe("Erlang_UnicodeUtil", () => {
         const cpResult = cp(improperInput);
 
         // cp/1 should return improper list: [97, bitstring("b"), bitstring("cd")]
-        assert.strictEqual(cpResult.data.length, 3, "cp/1 should return 3-element improper list");
-        assert.strictEqual(cpResult.isProper, false, "cp/1 result should be improper");
+        assert.strictEqual(
+          cpResult.data.length,
+          3,
+          "cp/1 should return 3-element improper list",
+        );
+        assert.strictEqual(
+          cpResult.isProper,
+          false,
+          "cp/1 result should be improper",
+        );
         assert.deepStrictEqual(
           cpResult,
           Type.improperList([
@@ -1033,11 +1041,14 @@ describe("Erlang_UnicodeUtil", () => {
         ]);
         const result = gc(improperInput);
 
-        assert.deepStrictEqual(result, Type.improperList([
-          Type.integer(97),
-          Type.bitstring("b"),
-          Type.bitstring("cd"),
-        ]));
+        assert.deepStrictEqual(
+          result,
+          Type.improperList([
+            Type.integer(97),
+            Type.bitstring("b"),
+            Type.bitstring("cd"),
+          ]),
+        );
       });
     });
 


bartblast:
@coderabbitai full review

assets/js/erlang/unicode_util.mjs
```diff
@@ -229,6 +229,180 @@ const Erlang_UnicodeUtil = {
   },
   // End cp/1
   // Deps: []
+  // Start gc/1
```
bartblast, line 703: Could you add a newline before `// Start gc/1` and remove `// Deps: [:unicode_util.cp/1]` (it's already defined correctly after the `// End gc/1` section)

assets/js/erlang/unicode_util.mjs
```diff
@@ -229,6 +229,180 @@ const Erlang_UnicodeUtil = {
...
+  // Deps: [:unicode_util.cp/1]
+  "gc/1": (arg) => {
+    const segmenter =
+      typeof Intl !== "undefined" && typeof Intl.Segmenter === "function"
```
bartblast: `Intl.Segmenter` has been [Baseline since April 2024](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) - supported in all modern browsers (Chrome 87+, Firefox 125+, Safari 15.4+) and Node.js 16+. The feature detection check and fallback path are dead code.

The fallback is also incorrect - it just returns the first codepoint, ignoring combining marks entirely - so if it ever *did* run, it would silently produce wrong results.

Also, a new `Intl.Segmenter` instance is created on every invocation of `gc/1`. I've added a shared instance at `ERTS.graphemeSegmenter`  please use that instead.

Please simplify to:

```javascript
const segmenter = ERTS.graphemeSegmenter;

const firstSegment = (text) => {
  const iterator = segmenter.segment(text)[Symbol.iterator]();
  const {done, value} = iterator.next();
  return done ? "" : value.segment;
};
```

assets/js/erlang/unicode_util.mjs
```diff
@@ -229,6 +229,180 @@ const Erlang_UnicodeUtil = {
...
+        // Keep collecting codepoints if boundary not yet determined
+        if (!boundaryKnown)
+          return collectCluster(tail, newCollectedText, newCollectedCodepoints);
+
+        // We've found the boundary. Check if we overshot (consumed too many codepoints).
+        const clusterLength = segmentLength;
+        const extraCount = consumedLength - clusterLength;
+        const adjustedTail =
+          extraCount > 0
+            ? prependCodepointToRest(
```
bartblast, line 873: When the algorithm overshoots the grapheme boundary, `prependCodepointToRest` is called with only the **last** collected codepoint.

If `extraCount > 1` (e.g., two non-combining characters were consumed before the boundary was detected), only the last one would be prepended back. The remaining extra codepoints would be silently dropped.

This should iterate in reverse over all extra codepoints, prepending each one, or collect all extra codepoints into a list to prepend in bulk. Consider adding a test case that exercises this path (e.g., a list mixing base characters and binaries that triggers multi-codepoint overshoot).

assets/js/erlang/unicode_util.mjs
```diff
@@ -229,6 +229,180 @@ const Erlang_UnicodeUtil = {
...
+      // (e.g., to check if the next character is a combining mark).
+      const collectCluster = (
+        remaining,
+        collectedText,
+        collectedCodepoints,
+      ) => {
+        const {codepoint, tail, error, empty} = extractHead(remaining);
+
+        if (error) return error;
+        if (empty) return Type.list();
```
bartblast: If `extractHead` returns `{empty: true}` after some codepoints have already been collected (i.e., `collectedCodepoints.length > 0`), this returns an empty list, discarding the collected codepoints.

This could happen with inputs like `[97, <<>>]` where an empty binary follows a base character - `cp/1` would eventually return an empty list, and the already-collected `97` would be lost.

Consider adding a guard:

```javascript
if (empty) {
  if (collectedCodepoints.length > 0) {
    const cluster = buildCluster(collectedCodepoints);
    return Type.list([cluster]);
  }

  return Type.list();
}
```

Either way, please add a test for `gc([97, <<>>])` to verify correct behavior and document the expected result

test/elixir/hologram/ex_js_consistency/erlang/unicode_util_test.exs
```diff
@@ -200,4 +200,66 @@ defmodule Hologram.ExJsConsistency.Erlang.UnicodeUtilTest do
...
+
+    test "handles list starting with binary" do
+      assert :unicode_util.gc(["ab", 98]) == [97, "b", 98]
+    end
+
+    test "handles binary with combining marks inside list" do
+      assert :unicode_util.gc(["e̊", 120]) == [[101, 778], "", 120]
+    end
+
+    test "raises FunctionClauseError for non-byte-aligned bitstring" do
```
bartblast, line 494: This test belongs to the "error handling" section. (same for the JS version)

test/elixir/hologram/ex_js_consistency/erlang/unicode_util_test.exs
```diff
@@ -200,4 +200,66 @@ defmodule Hologram.ExJsConsistency.Erlang.UnicodeUtilTest do
...
+      assert :unicode_util.gc(invalid_binary) == {:error, invalid_binary}
+    end
+
+    # Section: with list input
+
+    test "returns empty list for empty list" do
+      assert :unicode_util.gc([]) == []
+    end
+
+    test "handles list of integers" do
```
bartblast: The name "handles list of integers" doesn't convey what's being tested - that `gc/1` extracts a single-codepoint cluster from a list of non-combining codepoints and returns `[cluster | rest]`. A name like `"extracts single-codepoint cluster when next codepoint is non-combining"` would be clearer. The same applies to the corresponding JS test.

assets/js/erlang/unicode_util.mjs
mward-sudo: The grapheme clustering loop adds one codepoint at a time and marks the boundary when `segmentLength < consumedLength` (or other boundary conditions). Because consumed length only increases by one per iteration, the gap `consumedLength - segmentLength` is bounded to 1. That means we can overshoot by at most a single codepoint, so the earlier single-prepend logic already returns all codepoints. A multi-prepend path would be unreachable with the current algorithm.

bartblast: Got it, thanks!

bartblast:
@mward-sudo Could you resolve the branch conflicts?

bartblast:
@mward-sudo Could you check CodeRabbit's comments?

mward-sudo:
@coderabbitai review

bartblast:
Thank you @mward-sudo! 💜

PR #606, opened 2026-01-22
Port :elixir_utils.jaro_similarity/2 to JS

### LINKED ISSUE #569: Port: :elixir_utils.jaro_similarity/2

Petarj123:
closes #569 

## Summary by CodeRabbit

* **New Features**
  * Implemented Jaro similarity for measuring similarity between inputs in multiple formats (strings, character lists, integer lists), with normalization and robust error handling for invalid inputs.

* **Tests**
  * Added comprehensive JavaScript and Elixir test suites covering identical/empty/disjoint cases, transpositions, Unicode and emoji, cross-format consistency, nested/mixed inputs, and explicit error conditions.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
@@ -127,9 +127,11 @@ defmodule Elixir.Hologram.ExJsConsistency.Erlang.ElixirUtilsTest do
       test "raises ArgumentError for invalid UTF-8 bytes" do
         invalid_utf8 = <<255, 254, 253>>
 
-        assert_raise ArgumentError, fn ->
-          :elixir_utils.jaro_similarity(invalid_utf8, "test")
-        end
+        assert_error ArgumentError,
+                     "argument error: #{inspect(invalid_utf8)}",
+                     fn ->
+                       :elixir_utils.jaro_similarity(invalid_utf8, "test")
+                     end
       end
     end
   end


assets/js/erlang/elixir_utils.mjs
```diff
@@ -0,0 +1,130 @@
...
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const Erlang_Elixir_Utils = {
+  // Start jaro_similarity/2
+  "jaro_similarity/2": (str1, str2) => {
+    // Helper: Extract code points from a bitstring or list (including nested lists)
+    const extractCodePoints = (str) => {
+      if (Type.isBitstring(str)) {
+        Bitstring.maybeSetTextFromBytes(str);
+        return Array.from(str.text).map((c) => c.charCodeAt(0));
```
bartblast: Using `charCodeAt(0)` instead of `codePointAt(0)` causes incorrect results for characters outside the Basic Multilingual Plane (emoji, some CJK characters, etc.).

Example:
```javascript
"😀".charCodeAt(0)   // → 55357 (WRONG - returns high surrogate)
"😀".codePointAt(0)  // → 128512 (CORRECT - actual Unicode code point)
```

Could you  change all three occurrences:

```javascript
// Before:
return Array.from(str.text).map((c) => c.charCodeAt(0));
// After:
return Array.from(str.text).map((c) => c.codePointAt(0));
```

```javascript
// Before:
codePoints.push(n.text.charCodeAt(0));
// After:
codePoints.push(n.text.codePointAt(0));
```

```javascript
// Before:
codePoints.push(firstElem.text.charCodeAt(0));
// After:
codePoints.push(firstElem.text.codePointAt(0));
```

Also please add a test with emoji to verify this works correctly:
```javascript
it("handles emoji characters", () => {
  const result = jaroSimilarity(
    Type.bitstring("hello😀"),
    Type.bitstring("hello😀"),
  );

  assert.strictEqual(result.value, 1.0);
});
```

assets/js/erlang/elixir_utils.mjs
```diff
@@ -0,0 +1,130 @@
...
+
+// IMPORTANT!
+// If the given ported Erlang function calls other Erlang functions, then list such dependencies in the "Deps" comment (see :erlang./=/2 for an example).
+// Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
+
+const Erlang_Elixir_Utils = {
+  // Start jaro_similarity/2
+  "jaro_similarity/2": (str1, str2) => {
+    // Helper: Extract code points from a bitstring or list (including nested lists)
+    const extractCodePoints = (str) => {
```
bartblast, line 15: The `extractCodePoints` helper duplicates some logic that already exists in the ported `:unicode_util.cp/1` function (see `assets/js/erlang/unicode_util.mjs`).

The existing `cp/1` implementation handles several edge cases that `extractCodePoints` currently doesn't:

| Case | `cp/1` | `extractCodePoints` |
|------|--------|---------------------|
| Unicode codepoint range validation (0-0x10FFFF) | Yes | No |
| Surrogate pair detection (0xD800-0xDFFF) | Yes | No |
| Invalid UTF-8 bytes check | Yes | No |
| Non-byte-aligned bitstrings | Yes | No |
| Empty nested lists | Yes | Falls to error |

Could you please check if some of the inlined helpers in `unicode_util.cp/1` could be reused here? For example:
- `extractCodepointFromText`
- `isSurrogatePair`
- The validation logic

These helpers could potentially be extracted to a shared location (e.g., `Unicode` class or a new utility module) to avoid duplication and ensure consistent behavior. This doesn't need to block this PR, but it's worth considering for consistency.

At minimum, please verify that the current implementation produces the same results as the Elixir version for edge cases like:
- Invalid UTF-8 sequences
- Surrogate pairs in input
- Out-of-range code points

test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
```diff
@@ -0,0 +1,107 @@
...
+      test "handles slight deviations" do
+        assert_in_delta :elixir_utils.jaro_similarity("martha", "marhta"), 0.944, 0.001
+        assert_in_delta :elixir_utils.jaro_similarity("dwayne", "duane"), 0.822, 0.001
+        assert_in_delta :elixir_utils.jaro_similarity("dixon", "dicksonx"), 0.767, 0.001
+      end
+
+      test "returns 0.0 for completely different inputs" do
+        assert :elixir_utils.jaro_similarity("abc", "xyz") == 0.0
+      end
+
```
bartblast, line 25: We're missing a test for identical strings (longer than 1 character) returning `1.0`. Please add something like:

Elixir:
```elixir
test "returns 1.0 for identical strings" do
  assert :elixir_utils.jaro_similarity("hello", "hello") == 1.0
end
```

JS:
```javascript
it("returns 1.0 for identical strings", () => {
  const result = jaroSimilarity(
    Type.bitstring("hello"),
    Type.bitstring("hello"),
  );

  assert.strictEqual(result.value, 1.0);
});
```

test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
```diff
@@ -0,0 +1,107 @@
...
+        assert :elixir_utils.jaro_similarity(["a", "b", "c"], ["a", "b", "c"]) == 1.0
+      end
+
+      test "handles lists with mixed integers and strings" do
+        assert :elixir_utils.jaro_similarity([97, "b", 99], [97, "b", 99]) == 1.0
+      end
+
+      test "handles lists with multi-character strings" do
+        result = :elixir_utils.jaro_similarity(["ab", "cd"], ["ab", "cd"])
+        assert result == 1
```
bartblast: `result == 1.0`

test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
```diff
@@ -0,0 +1,107 @@
...
+      end
+
+      test "handles nested lists" do
```
bartblast, line 93: The code points `1` and `2` are control characters (SOH and STX). While this technically works, using printable ASCII code points would be clearer and more consistent with similar tests.

This could be:
```elixir
test "handles nested lists" do
  # Using ASCII code points for 'a' (97), 'b' (98), 'c' (99)
  result = :elixir_utils.jaro_similarity([97, 98, [99]], [97, 98, [99]])

  assert result == 1.0
end
```

(and update the corresponding JS test accordingly)

test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
```diff
@@ -0,0 +1,107 @@
...
+        assert_in_delta :elixir_utils.jaro_similarity("martha", "marhta"), 0.944, 0.001
+        assert_in_delta :elixir_utils.jaro_similarity("dwayne", "duane"), 0.822, 0.001
+        assert_in_delta :elixir_utils.jaro_similarity("dixon", "dicksonx"), 0.767, 0.001
+      end
+
+      test "returns 0.0 for completely different inputs" do
+        assert :elixir_utils.jaro_similarity("abc", "xyz") == 0.0
+      end
+
+      test "handles empty inputs" do
```
bartblast: This test verifies multiple scenarios in a single test. Splitting it would provide better error isolation:

Example:
```javascript
it("returns 1.0 when both inputs are empty", () => { ... });
it("returns 0.0 when first input is empty", () => { ... });
it("returns 0.0 when second input is empty", () => { ... });
```

(remember about Elixir version)

test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
```diff
@@ -0,0 +1,107 @@
...
+        assert :elixir_utils.jaro_similarity("", "") == 1.0
+        assert :elixir_utils.jaro_similarity("", "hello") == 0.0
+        assert :elixir_utils.jaro_similarity("hello", "") == 0.0
+      end
+
+      test "handles single character inputs" do
```
bartblast: Please split into separate test cases. (remember about JS also)

test/elixir/hologram/ex_js_consistency/erlang/elixir_utils_test.exs
```diff
@@ -0,0 +1,107 @@
...
+        # Known issue in :elixir_utils.jaro_similarity/2
+        # will be fixed when Elixir requires Erlang/OTP 27+
+        # and switches to :string.jaro_similarity/2
+        assert :elixir_utils.jaro_similarity("a", "a") == 0.0
+        assert :elixir_utils.jaro_similarity("a", "b") == 0.0
+      end
+
+      test "handles transpositions" do
```
bartblast: Please split into separate test cases. (remember about JS also)

test/javascript/erlang/elixir_utils_test.mjs
```diff
@@ -0,0 +1,224 @@
...
+      const intResult = jaroSimilarity(
+        Type.list([Type.integer(97), Type.integer(98), Type.integer(99)]),
+        Type.list([Type.integer(97), Type.integer(98), Type.integer(100)]),
+      );
+
+      assert.deepStrictEqual(strResult, charResult);
+      assert.deepStrictEqual(charResult, intResult);
+    });
+
+    it("handles slight deviations", () => {
```
bartblast: The purpose of consistency tests is to verify that the JS implementation produces exactly the same results as the Elixir implementation. Using `assert_in_delta` defeats this purpose - it could mask precision bugs where JS and Elixir produce slightly different values.

Please use exact values instead:

```elixir
# Before (hides potential differences):
assert_in_delta :elixir_utils.jaro_similarity("martha", "marhta"), 0.944, 0.001

# After (verifies exact match):
assert :elixir_utils.jaro_similarity("martha", "marhta") == 0.9444444444444445
assert :elixir_utils.jaro_similarity("dwayne", "duane") == 0.8222222222222223
assert :elixir_utils.jaro_similarity("dixon", "dicksonx") == 0.7666666666666666
```

(and update the corresponding JS tests)

assets/js/erlang/elixir_utils.mjs
Petarj123: Hi @bartblast,
I agree that we should use :unicode_util.cp/1, but current port does not raise same exceptions as erlang does. Should we adjust port of :unicode_util.cp/1 first, my handrolled approach works, but i feel like this duplication is not necessary. What do you think? 
Screenshot_20260127_224223
Screenshot_20260127_224255

bartblast: @mward-sudo Could you take a look at this and share your thoughts on how difficult it would be to reproduce this behavior in cp/1 (and whether it even makes sense from your perspective)?

mward-sudo: > @mward-sudo Could you take a look at this and share your thoughts on how difficult it would be to reproduce this behavior in cp/1 (and whether it even makes sense from your perspective)?

@bartblast @Petarj123 If it's acting out of spec then raise a bug report detailing the issue and I can take a look at a fix.

Put the code that isn't complying as text in the bug.

bartblast: @Petarj123 Could you create a new GitHub issue that details the discrepancies you see in the Erlang vs JS implementation of `:unicode_util.cp/1`? Please include the non-complying code as text (as @mward-sudo requested).

I don't worry about the duplication at this stage - we can always refactor after v0.7.0. What worries me is that `:unicode_util.cp/1` handles edge cases (Unicode range validation, surrogate pair detection, invalid UTF-8 bytes, non-byte-aligned bitstrings, empty nested lists) that `extractCodePoints` currently doesn't.

If @mward-sudo can fix `cp/1` quickly, we'd use it here instead of the custom helper. Otherwise we can proceed as-is and revisit later.

Petarj123: @bartblast Do we just go with :unicode_util.cp/1 regardless? Once this small issue is fixed, everything should work fine and we won’t have to go back and address a parallel implementation.

bartblast: @Petarj123 If we switch `extractCodePoints` to use `:unicode_util.cp/1` now, the tests will fail in CI. As reported in #636, the JS port of `cp/1` currently diverges from Erlang behaviour: when given a list with multiple elements where at least one is invalid, Erlang raises `no function clause matching in :unicode_util.cpl/2` (referencing the internal `cpl/2` helper), but our JS port raises `no function clause matching in :unicode_util.cp/1` instead. Any test that expects the correct Erlang error would fail, and the tests in this PR depend on that behaviour. So we have two options:

1. Fix `cp/1` first to match the Erlang error behaviour, then refactor `extractCodePoints` to use it here.
2. Merge this PR as-is and refactor `extractCodePoints` to use `cp/1` in a follow-up once #636 is resolved.

Which do you prefer?

mward-sudo: @bartblast I'm doing a quite substantial rewrite of cp as its current form is not handling edge cases at all well. I expect to have it done reasonably soon, but it might be prudent to land the current code as it is and refactor it once my rewrite lands. Would be good to note with a TODO comment where the logic would be replaced.

mward-sudo: #637 reimplements `:unicode_util.cp/1 `for spec compliant behaviour.

bartblast: @Petarj123 The function has been fixed by @mward-sudo, it's ready to be used in your port implementation.

Petarj123: @bartblast Fixed it, but I also found out that :elixir_utils.jaro_similarity/2 returns ArgumentError without a specific argument position when an invalid UTF-8 byte sequence is passed. However, Interpreter.buildArgumentErrorMsg expects an argumentIndex parameter. I don't know if this is important at this time, but it is a divergence from Erlang.
e

bartblast: @Petarj123 The helper builds messages for the most common ArgumentError cases, but for this one we need to use something like this:

```js
Interpreter.raiseArgumentError(`argument error: ${Interpreter.inspect(arg)}`)
```

bartblast:
@Petarj123 Could you revert changes in `.tool-versions`?

Petarj123:
> @Petarj123 Could you revert changes in `.tool-versions`?

Done

bartblast:
@Petarj123 One more thing I've noticed:
```elixir
test "raises ArgumentError for invalid UTF-8 bytes" do
  invalid_utf8 = <<255, 254, 253>>

  assert_raise ArgumentError, fn ->
    :elixir_utils.jaro_similarity(invalid_utf8, "test")
  end
end
```

This one uses `assert_raise` instead of `assert_error` - we need to assert on the error message here as well.

Petarj123:
> @Petarj123 One more thing I've noticed:
> 
> ```elixir
> test "raises ArgumentError for invalid UTF-8 bytes" do
>   invalid_utf8 = <<255, 254, 253>>
> 
>   assert_raise ArgumentError, fn ->
>     :elixir_utils.jaro_similarity(invalid_utf8, "test")
>   end
> end
> ```
> 
> This one uses `assert_raise` instead of `assert_error` - we need to assert on the error message here as well.

Done

bartblast:
Looks good, thank you for you help @Petarj123! 💜

Issue #557, opened 2026-01-16
Port :sets.is_disjoint/2 to JS


bartblast:

Lucassifoni:
@bartblast I'm up for it if you still need help !

bartblast:
Go ahead! :)

bartblast:
Hey @Lucassifoni, I'm planning to release all the ported functions in a few days. Any chance you'll be able to wrap this one up soon?

Lucassifoni:
Sure, I’ll push it this morning !

PR #612, opened 2026-01-23
Port :sets.is_disjoint/2 to JS

Lucassifoni:
A few notes : 

I wanted to use JS's Set API to hopefully leverage an efficient implementation in engines. But it uses referential equality, so `(new Set([{a: 5}]).isDisjointFrom((new Set([{a: 5}])) === true`. So I kept implementing with JS's Set but working on sets of encoded map keys - is that alright ?

Edit : not alright, lacks browser support, I've rewritten. Linked to issue #557 .

## Summary by CodeRabbit

* **New Features**
  * Added is_disjoint/2 to check whether two sets share no common elements, with input validation.

* **Tests**
  * Added comprehensive tests for is_disjoint/2 covering primitives and objects, empty-set interactions, identical sets, strict type matching, and error cases when inputs are not valid sets.

### DIFF
--- a/test/javascript/erlang/sets_test.mjs
+++ b/test/javascript/erlang/sets_test.mjs
@@ -146,8 +146,6 @@ describe("Erlang_Sets", () => {
     });
   });
 
-
-
   describe("filter/2", () => {
     const filter_2 = Erlang_Sets["filter/2"];
 


assets/js/erlang/sets.mjs
```diff
@@ -90,6 +90,26 @@ const Erlang_Sets = {
...
+            set1,
+            set2,
+          ]),
+        );
+      }
+    });
+
+    const encodedKeys1 = new Set(Object.keys(set1.data));
+    const encodedKeys2 = new Set(Object.keys(set2.data));
+    return Type.boolean(encodedKeys1.isDisjointFrom(encodedKeys2));
```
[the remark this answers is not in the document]
Lucassifoni: My bad, I thought support was a lot wider. I'm rewriting without it.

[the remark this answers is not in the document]
Lucassifoni: I agree with the robot that iterating the smaller set checking for keys in the larger is the way to go... so I've rewritten using has()

Lucassifoni:
Looks like the error is inconsistent between ubuntu-latest elixir 1.18.2 / otp 27.2 and the same versions on windows-200. I'll investigate further. An unrelated test timed out on windows-2022 1.18.2 / otp 25.3, I'll see if this one comes up again or not.

assets/js/erlang/sets.mjs
```diff
@@ -90,6 +90,38 @@ const Erlang_Sets = {
   // End del_element/2
   // Deps: [:maps.remove/2]
 
+  // Start is_disjoint/2
+  "is_disjoint/2": (set1, set2) => {
+    [set1, set2].forEach((set) => {
```
bartblast: Could you replace the forEach with a direct check:
```javascript
if (!Type.isMap(set1) || !Type.isMap(set2)) {
  Interpreter.raiseFunctionClauseError(
    Interpreter.buildFunctionClauseErrorMsg(":sets.is_disjoint/2", [
      set1,
      set2,
    ]),
  );
}
```

This avoids allocating an array and benefits from short-circuit evaluation. It'll also minify well once we optimize bundle size.

test/elixir/hologram/ex_js_consistency/erlang/sets_test.exs
```diff
@@ -103,6 +103,82 @@ defmodule Hologram.ExJsConsistency.Erlang.SetsTest do
...
+
+    test "uses strict matching (integer vs float)" do
+      set_int = :sets.from_list([1], version: 2)
+      set_float = :sets.from_list([1.0], version: 2)
+
+      assert :sets.is_disjoint(set_int, set_float) == true
+    end
+
+    test "raises FunctionClauseError if the first argument is not a set", %{set_123: set_123} do
+      expected_msg = ~r/no function clause matching in :sets\.(is_disjoint\/2|size\/1).*:abc/s
```
bartblast: Could you use `build_function_clause_error_msg/1` in such cases? If OTP versions differ in their error messages, I'd prefer adding a version conditional that compares against the appropriate message for each version. If the output contains non-deterministic parts (like anonymous function references), I'd prefer specifying the full error message structure with regex only for those variable parts. This ensures we catch any unexpected changes in Erlang/Elixir error formatting.

The current regex allows both `:sets.is_disjoint/2` and `:sets.size/1`, but the JS implementation always raises with `:sets.is_disjoint/2`. If the reported function depends on OTP version, use the latest OTP version as the reference for the JS implementation. However, please still include both variants in the Elixir tests (with version conditionals) so we have a good overview of the overall situation across versions.

assets/js/erlang/sets.mjs
```diff
@@ -90,6 +90,38 @@ const Erlang_Sets = {
   // End del_element/2
   // Deps: [:maps.remove/2]
 
+  // Start is_disjoint/2
+  "is_disjoint/2": (set1, set2) => {
```
bartblast: Could you order alphabetically (here and in the test files).

assets/js/erlang/sets.mjs
```diff
@@ -90,6 +90,38 @@ const Erlang_Sets = {
...
+
+    const encodedKeys1 = new Set(Object.keys(set1.data));
+    const encodedKeys2 = new Set(Object.keys(set2.data));
+
+    const [smaller, larger] =
+      encodedKeys1.size <= encodedKeys2.size
+        ? [encodedKeys1, encodedKeys2]
+        : [encodedKeys2, encodedKeys1];
+
+    for (const key of smaller) {
```
bartblast: I noticed your comment in the PR description about `Set.prototype.isDisjointFrom()` lacking browser support. It's actually been available in all major browsers since June 2024 (Baseline 2024) - that's 1.5 years now. You could simplify the implementation to:

```javascript
const encodedKeys1 = new Set(Object.keys(set1.data));
const encodedKeys2 = new Set(Object.keys(set2.data));

return Type.boolean(encodedKeys1.isDisjointFrom(encodedKeys2));
```

This leverages the browser's native implementation which is likely more optimized than a manual loop.

Generally, if a feature is marked as Baseline (even "Newly Available") on MDN, that's a green flag for us to use it:
- Baseline means all major browser engines (Chromium, Firefox, WebKit) support it
- "Newly Available" still requires 6+ months of cross-browser stability
- Modern browsers auto-update, so users get new features quickly
- Native implementations are typically faster and better optimized than JS equivalents
- Manual implementations add maintenance burden and potential for bugs

Lucassifoni: Perfect ! The low-ish score on CanIUse.com made me think it would be wiser to avoid it but I agree with your arguments (and prefer it). I'll be able to make all the changes you requested tomorrow morning.

Lucassifoni:
I read the OTP source code for Sets and did not find a conclusive reason to have the error emitted from is_disjoint instead of size when the arguments are not maps. I do not know what happened in the specific CI run that failed but it does not now, while I've normalized the errors messages to size.

bartblast:
Thank you Lucas! 👍

PR #617, opened 2026-01-24
Port :binary.match/2 and :binary.match/3 to JS

### LINKED ISSUE #380: Port :binary.match/2 and :binary.match/3 to JS

mward-sudo:
Closes #380 

### Depends upon 

- [x] #616 

Merge only after the above PR is merged, may require rebasing.

## Summary by CodeRabbit

* **New Features**
  * Added public binary match APIs (two- and three-argument forms) with scope-aware matching and compiled-pattern support.

* **Improvements**
  * Prefer longest/earliest match; tightened termination and scope semantics (null denotes "until end"); stricter option/argument validation; clearer error positions; consistent trimming/UTF‑8 handling; improved routing for pattern types.

* **Tests**
  * Extensive Elixir and JS tests covering match/split behavior, scoped ranges, compiled patterns, overlaps, and many error scenarios.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
@@ -488,7 +488,7 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
       end
     end
 
-    test "raises ArgumentError with negative scope length" do
+    test "raises ArgumentError when scope start plus negative length is below zero" do
       assert_raise ArgumentError, fn ->
         :binary.match("test", "es", scope: {0, -1})
       end
--- a/test/javascript/erlang/binary_test.mjs
+++ b/test/javascript/erlang/binary_test.mjs
@@ -986,7 +986,7 @@ describe("Erlang_Binary", () => {
         );
       });
 
-      it("raises ArgumentError with negative scope length", () => {
+      it("raises ArgumentError when scope start plus negative length is below zero", () => {
         const subject = Bitstring.fromText("test");
         const pattern = Bitstring.fromText("es");
         const options = Type.list([


mward-sudo:
@coderabbitai full review

assets/js/erlang/binary.mjs
```diff
@@ -268,119 +434,281 @@ const Erlang_Binary = {
...
+    const scopeEnd = Math.min(
+      scopeStart + effectiveLength,
+      subject.bytes.length,
+    );
+
+    // No search range available
+    if (scopeStart >= subject.bytes.length || scopeLength === 0) {
+      const parts = [bytesToBitstring(subject.bytes)];
+      return Type.list(applyTrim(parts));
+    }
```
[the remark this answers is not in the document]
mward-sudo: This seems to be outside of the scope of this issue

bartblast:
@mward-sudo FYI I've made some changes to the split functions and I'm adding more tests. This may affect this PR. I'll give you a heads up soon when I'm done.

bartblast:
@mward-sudo I've finished updating split functions related code. Could you pull from dev?

mward-sudo:
Tired out the bunny :(

mward-sudo:
@coderabbitai full review

mward-sudo:
@coderabbitai review

assets/js/erlang/binary.mjs
```diff
@@ -570,13 +711,17 @@ const Erlang_Binary = {
...
 
+    // For negative scopeLength, ensure slice bounds are in correct order
+    const actualStart = Math.min(scopeStart, scopeEnd);
+    const actualEnd = Math.max(scopeStart, scopeEnd);
+
     // No search range available
     if (scopeStart >= subject.bytes.length || scopeLength === 0) {
       const parts = [bytesToBitstring(subject.bytes)];
       return Type.list(applyTrim(parts));
     }
```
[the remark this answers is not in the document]
mward-sudo: Out of scope for this PR. Will address in a future PR.

bartblast:
@mward-sudo Since #619 depends on this one (#617), I should review this one first, right?

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -286,6 +286,226 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
     end
   end
 
+  describe "match/2" do
```
bartblast, line 289: The `match/2` function simply delegates to `match/3`:
```javascript
"match/2": (subject, pattern) => {
  return Erlang_Binary["match/3"](subject, pattern, Type.list());
},
```

Following the established pattern (see `split/2` which has only 1 test), `match/2` should have only a single delegation test. The other `match/2` tests are redundant since they're all testing `match/3` indirectly.

Could you remove all `match/2` tests except one that verifies delegation (similar to how `split/2` is tested).

Some of these test should be probably moved to `match/3`, because we're missing coverage there, e.g. "raises ArgumentError if subject is a non-binary bitstring" (please verify).

test/javascript/erlang/binary_test.mjs
```diff
@@ -493,6 +493,550 @@ describe("Erlang_Binary", () => {
...
+      ]);
+
+      assertBoxedError(
+        () => match(subject, pattern, options),
+        "ArgumentError",
+        Interpreter.buildArgumentErrorMsg(3, "invalid options"),
+      );
+    });
+
+    it("raises ArgumentError with negative scope length", () => {
```
bartblast: This test name is misleading. According to the Erlang docs, negative scope lengths ARE allowed when `Start + Length >= 0`. The test uses `{0, -1}` which fails because `0 + (-1) = -1 < 0`.

Suggested rename:
```javascript
it("raises ArgumentError when scope start plus negative length is below zero", () => {
```

Same applies to the corresponding Elixir test.

assets/js/erlang/binary.mjs
bartblast, line 112: Since the PR changes `_parse_search_opts/2` to allow negative scope lengths, add a test verifying `split/3` works with valid negative scope lengths (to match the `match/3` test "accepts negative scope length (reverse part)"):

```elixir
test "accepts negative scope length (reverse part)" do
  subject = "hello world"
  opts = [scope: {byte_size(subject), -5}]
  result = :binary.split(subject, " ", opts)

  assert result == ["hello", "world"]
end
```

```javascript
it("accepts negative scope length (reverse part)", () => {
  const subject = Bitstring.fromText("hello world");
  const pattern = Bitstring.fromText(" ");

  const opts = Type.list([
    Type.tuple([
      Type.atom("scope"),
      Type.tuple([Type.integer(11), Type.integer(-5)]),
    ]),
  ]);

  const result = split(subject, pattern, options);

  assertBoxedStrictEqual(
    result,
    Type.list([Bitstring.fromText("hello"), Bitstring.fromText("world")]),
  );
});
```

mward-sudo: There's a bug in my split/3 function so this test fail.

I believe that I've fixed the bug in the follow up refactoring pull request, but I've logged it as an issue so that I'll remember to check. I'll also ensure that this test or equivalent is present.

bartblast:
Thank you @mward-sudo! 🚀

PR #629, opened 2026-01-27
Port :binary.replace/3 and :binary.replace/4 to JS

### LINKED ISSUE #382: Port :binary.replace/3 and :binary.replace/4 to JS

mward-sudo:
Closes #382 

## Summary by CodeRabbit

* **New Features**
  * Adds a full-featured binary replacement API plus a convenience wrapper, supporting pattern lists, scoped regions, global mode, function-based replacements, and precompiled pattern formats (BM/AC).

* **Tests**
  * Adds extensive tests covering functionality, options, error handling, scope and global interactions, function-based replacements, compiled-pattern behavior, and numerous edge cases.

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
@@ -1074,6 +1074,10 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
       assert result == "hello world"
     end
 
+    test "supports negative scope length within range" do
+      assert :binary.replace("test", "t", "x", scope: {1, -1}) == "xest"
+    end
+
     # With compiled pattern
 
     test "works with compiled Boyer-Moore pattern" do
@@ -1120,6 +1124,12 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
                    {:binary, :replace, ["test", "es", "x", :invalid]}
     end
 
+    test "raises ArgumentError for improper list options" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(4, "invalid options"),
+                   {:binary, :replace, ["test", "t", "x", [:global | :bad]]}
+    end
+
     test "raises ArgumentError when scope start exceeds subject length" do
       assert_error ArgumentError,
                    build_argument_error_msg(4, "invalid options"),
@@ -1130,6 +1140,12 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
       assert_error ArgumentError,
                    build_argument_error_msg(4, "invalid options"),
                    {:binary, :replace, ["test", "st", "x", [scope: {0, 100}]]}
+    end
+
+    test "raises ArgumentError when scope start plus negative length is below zero" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(4, "invalid options"),
+                   {:binary, :replace, ["test", "t", "x", [scope: {0, -1}]]}
     end
 
     test "raises ArgumentError when replacement is an atom" do
--- a/test/javascript/erlang/binary_test.mjs
+++ b/test/javascript/erlang/binary_test.mjs
@@ -2523,6 +2523,22 @@ describe("Erlang_Binary", () => {
 
         assertBoxedStrictEqual(result, subject);
       });
+
+      it("supports negative scope length within range", () => {
+        const subject = Bitstring.fromText("test");
+        const pattern = Bitstring.fromText("t");
+        const replacement = Bitstring.fromText("x");
+        const options = Type.list([
+          Type.tuple([
+            Type.atom("scope"),
+            Type.tuple([Type.integer(1), Type.integer(-1)]),
+          ]),
+        ]);
+
+        const result = replace(subject, pattern, replacement, options);
+
+        assertBoxedStrictEqual(result, Bitstring.fromText("xest"));
+      });
     });
 
     describe("with compiled pattern", () => {
@@ -2643,6 +2659,22 @@ describe("Erlang_Binary", () => {
         );
       });
 
+      it("raises ArgumentError for improper list options", () => {
+        const subject = Bitstring.fromText("test");
+        const pattern = Bitstring.fromText("t");
+        const replacement = Bitstring.fromText("x");
+        const options = Type.improperList([
+          Type.atom("global"),
+          Type.atom("bad"),
+        ]);
+
+        assertBoxedError(
+          () => replace(subject, pattern, replacement, options),
+          "ArgumentError",
+          Interpreter.buildArgumentErrorMsg(4, "invalid options"),
+        );
+      });
+
       it("raises ArgumentError when scope extends beyond subject", () => {
         const subject = Bitstring.fromText("test");
         const pattern = Bitstring.fromText("st");
@@ -2651,6 +2683,24 @@ describe("Erlang_Binary", () => {
           Type.tuple([
             Type.atom("scope"),
             Type.tuple([Type.integer(0), Type.integer(100)]),
+          ]),
+        ]);
+
+        assertBoxedError(
+          () => replace(subject, pattern, replacement, options),
+          "ArgumentError",
+          Interpreter.buildArgumentErrorMsg(4, "invalid options"),
+        );
+      });
+
+      it("raises ArgumentError when scope start plus negative length is below zero", () => {
+        const subject = Bitstring.fromText("test");
+        const pattern = Bitstring.fromText("t");
+        const replacement = Bitstring.fromText("x");
+        const options = Type.list([
+          Type.tuple([
+            Type.atom("scope"),
+            Type.tuple([Type.integer(0), Type.integer(-1)]),
           ]),
         ]);
 


test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -753,4 +753,301 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+      assert :binary.replace("hello-world_test", ["-", "_"], "X", [:global]) ==
+               "helloXworldXtest"
+    end
+
+    test "handles consecutive patterns with :global" do
+      assert :binary.replace("a--b--c", "-", "X", [:global]) == "aXXbXXc"
+    end
+
+    test "handles empty replacement with :global" do
+      assert :binary.replace("a1b2c3", "1", "", [:global]) == "ab2c3"
```
bartblast: This test doesn't effectively demonstrate the `:global` option because the subject `"a1b2c3"` only contains one occurrence of the pattern `"1"`. The result `"ab2c3"` would be the same with or without `:global`.

Please use a subject with multiple occurrences of the pattern, e.g.:

```elixir
test "handles empty replacement with :global" do
  assert :binary.replace("a1b1c1", "1", "", [:global]) == "abc"
end
```

(same for the corresponding JS test)

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -753,4 +753,301 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+          "x",
+          [:global, {:insert_replaced, 1}]
+        )
+
+      expected_binary = "xabxaxabxa"
+      expected_positions = [0, 2, 3, 5]
+
+      assert result == expected_binary
+
+      assert {result, expected_positions} ==
```
bartblast, line 1015: The second assertion testing `:binary.matches` positions is testing a different function and adds complexity without verifying `replace/4` behavior. The first assertion (`assert result == expected_binary`) is sufficient. Please simplify to just:

```elixir
test "works with :global and insert_replaced" do
  assert :binary.replace("abaaba", "a", "x", [:global, {:insert_replaced, 1}]) == "xabxaxabxa"
end
```

test/javascript/erlang/binary_test.mjs
```diff
@@ -1700,4 +1700,701 @@ describe("Erlang_Binary", () => {
...
+            Type.list([Type.integer(1), Type.integer(1)]),
+          ]),
+        ]);
+
+        const result = replace(subject, pattern, replacement, options);
+
+        assertBoxedStrictEqual(result, Bitstring.fromText("ab[cdcd]e"));
+      });
+
+      it("works with :global and insert_replaced", () => {
```
bartblast, line 2339: This test is not consistent with the JS version.

test/javascript/erlang/binary_test.mjs
```diff
@@ -1700,4 +1700,701 @@ describe("Erlang_Binary", () => {
...
+            Type.tuple([Type.integer(8), Type.integer(3)]),
+          ]),
+        ]);
+
+        const result = replace(subject, pattern, replacement, options);
+
+        assertBoxedStrictEqual(result, Bitstring.fromText("abc def X"));
+      });
+
+      it("returns original subject when pattern outside scope", () => {
```
bartblast: The test name is misleading. The scope `{0, 3}` on subject `"abc def abc"` with pattern `"abc"` **does** match (the first "abc" is at position 0-2, within scope 0-3), so it replaces and returns `"X def abc"`. This is not "returns original subject when pattern outside scope" -- it's actually testing that replacement works within a narrow scope.

Could you rename it to: `"replaces pattern within scope"` (which is what the Elixir counterpart is named).

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -753,4 +753,301 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+                   build_argument_error_msg(4, "invalid options"),
+                   {:binary, :replace, ["hello", "l", "[]", [insert_replaced: [1, -1]]]}
+    end
+
+    test "raises ArgumentError when scope start is negative" do
+      assert_error ArgumentError,
+                   build_argument_error_msg(4, "invalid options"),
+                   {:binary, :replace, ["hello", "l", "X", [scope: {-1, 5}]]}
+    end
+
```
bartblast, line 1216: We're missing tests that exist in the `split/3` suite:

1. **Negative scope length**: Add a test for `scope: {2, -1}` or similar. The `split/3` tests include this scenario (`scope: {0, -1}`).

2. **Improper list as options**: Add a test for `:binary.replace("test", "t", "x", [:global | :bad])`. The `split/3` tests include `"raises ArgumentError for improper list options"`.

Both should raise `ArgumentError` with `build_argument_error_msg(4, "invalid options")`.

(Same for JS tests)

test/javascript/erlang/binary_test.mjs
```diff
@@ -1700,4 +1700,701 @@ describe("Erlang_Binary", () => {
...
+        };
+        const options = Type.list([Type.atom("global")]);
+
+        const result = replace(subject, pattern, replacement, options);
+
+        assert.equal(callCount, 2);
+        assertBoxedStrictEqual(result, Bitstring.fromText("XcXc"));
+      });
+
+      it("function receives matched binary", () => {
```
bartblast, line 2411: This one is redundant IMO.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
```diff
@@ -753,4 +753,301 @@ defmodule Hologram.ExJsConsistency.Erlang.BinaryTest do
...
+               "hello [world]"
+    end
+
+    test "calls function for each match with :global" do
+      result = :binary.replace("abcabc", "ab", fn _m -> "X" end, [:global])
+
+      assert result == "XcXc"
+    end
+
+    test "function receives matched binary" do
```
bartblast, line 1035: This one is redundant IMO.

assets/js/erlang/binary.mjs
```diff
@@ -811,6 +811,433 @@ const Erlang_Binary = {
...
+        : [Number(insertPositions.value)];
+
+      Bitstring.maybeSetBytesFromText(replacementBin);
+      Bitstring.maybeSetBytesFromText(matchedBin);
+
+      const repBytes = replacementBin.bytes;
+      const replacementLength = repBytes.length;
+
+      for (const pos of positions) {
+        if (pos > replacementLength) {
```
bartblast, line 1032: The Erlang docs say: "If any position specified in InsPos is greater than the size of the replacement binary, a badarg exception is raised.

Is this behaviour correct?

lib/hologram/compiler/call_graph.ex
```diff
@@ -28,6 +28,11 @@ defmodule Hologram.Compiler.CallGraph do
     {{:binary, :split, 3}, {:binary, :_parse_search_opts, 2}},
     {{:binary, :split, 3}, {:binary, :compile_pattern, 1}},
     {{:binary, :split, 3}, {:binary, :match, 3}},
+    {{:binary, :replace, 3}, {:binary, :replace, 4}},
+    {{:binary, :replace, 4}, {:binary, :compile_pattern, 1}},
+    {{:binary, :replace, 4}, {:binary, :match, 3}},
+    {{:binary, :replace, 4}, {:binary, :split, 3}},
+    {{:binary, :replace, 4}, {:erlang, :iolist_to_binary, 1}},
```
bartblast, line 39: Could you order this entry alphabetically?

test/javascript/erlang/binary_test.mjs
mward-sudo: AI suggests:

It uniquely asserts that the replacement callback is invoked with the actual matched binary (verifying the argument passed in), which is not covered by the other callback tests that only check return handling or call counts.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
mward-sudo: AI suggests:

"It’s still needed. That test is the only one that checks the callback receives the exact matched binary (and not something else), while the other callback tests only cover call counts or return handling. It guards against regressions where the replacement function might be called with the wrong argument."

assets/js/erlang/binary.mjs
mward-sudo: The CLI confirms BEAM raises on that case:

```
elixir -e ':binary.replace("abc","b","x", insert_replaced: 2)'

** (ArgumentError) 4th argument: invalid options
```

This matches the docs: a position greater than the replacement length triggers `badarg` (surfacing as `ArgumentError`).

lib/hologram/compiler/call_graph.ex
mward-sudo: Already alphabetical. 

ordered by module first (3 `:binary`, followed by 1 `:erlang`), the 3 `:binary` module functions alphabetical: c -> m -> s

test/javascript/erlang/binary_test.mjs
bartblast: @mward-sudo The "uses function for replacement" test already verifies this. It uses matched.text in the transformation ("[" + matched.text + "]"), so if the function received the wrong argument, the result wouldn't be "hello [world]" and the test would fail.

test/elixir/hologram/ex_js_consistency/erlang/binary_test.exs
bartblast: The "uses function for replacement" test already verifies this. It uses matched.text in the transformation ("[" + matched.text + "]"), so if the function received the wrong argument, the result wouldn't be "hello [world]" and the test would fail.

bartblast:
Thank you @mward-sudo! 👍 It's good to merge (I'll remove the test case that is redundant in my opinion, see my reasoning:  https://github.com/bartblast/hologram/pull/629#discussion_r2738464487).

PR #634, opened 2026-01-27
Port :os.system_time/0 to JS

### LINKED ISSUE #609: Port :os.system_time/0 to JS
```erl
-spec system_time() -> integer().
```

- [Erlang spec docs](https://www.erlang.org/doc/apps/kernel/os.html#system_time/0)

mward-sudo:
Closes #609 

## Summary by CodeRabbit

* **New Features**
  * Added a public system_time/0 function that returns the current system time in nanoseconds for high-precision timing.

* **Tests**
  * Added cross-language tests confirming the function returns an integer nanosecond timestamp within the observed wall-clock window (pre/post invocation bounds).

### DIFF
--- a/test/elixir/hologram/ex_js_consistency/erlang/os_test.exs
+++ b/test/elixir/hologram/ex_js_consistency/erlang/os_test.exs
@@ -10,7 +10,7 @@ defmodule Hologram.ExJsConsistency.Erlang.OsTest do
   @moduletag :consistency
 
   describe "system_time/0" do
-    test "returns current system time in nanoseconds" do
+    test "returns current system time in native time unit (nanoseconds)" do
       before_ms = System.os_time(:millisecond)
       result = :os.system_time()
       after_ms = System.os_time(:millisecond)
@@ -18,14 +18,6 @@ defmodule Hologram.ExJsConsistency.Erlang.OsTest do
       assert is_integer(result)
       assert result >= before_ms * 1_000_000
       assert result <= (after_ms + 1) * 1_000_000
-    end
-
-    test "returns different values on subsequent calls" do
-      result1 = :os.system_time()
-      result2 = :os.system_time()
-
-      assert is_integer(result1)
-      assert is_integer(result2)
     end
   end
 
--- a/test/javascript/erlang/os_test.mjs
+++ b/test/javascript/erlang/os_test.mjs
@@ -18,45 +18,14 @@ describe("Erlang_Os", () => {
   describe("system_time/0", () => {
     const systemTime = Erlang_Os["system_time/0"];
 
-    it("returns current system time in nanoseconds", () => {
+    it("returns current system time in native time unit (nanoseconds)", () => {
       const beforeMs = Date.now();
       const result = systemTime();
       const afterMs = Date.now();
 
-      assert.ok(Type.isInteger(result));
-      assert.ok(result.value >= BigInt(beforeMs) * 1_000_000n);
-      assert.ok(result.value <= BigInt(afterMs + 1) * 1_000_000n);
-    });
-
-    it("returns positive values", () => {
-      const result = systemTime();
-
-      assert.ok(result.value > 0n);
-    });
-
-    it("converts correctly to milliseconds", () => {
-      const result = systemTime();
-      const resultMs = Number(result.value / 1_000_000n);
-      const currentMs = Date.now();
-
-      // Allow 100ms tolerance for test execution time
-      assert.ok(Math.abs(resultMs - currentMs) < 100);
-    });
-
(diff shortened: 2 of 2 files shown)


assets/js/erlang/os.mjs
```diff
@@ -7,6 +7,15 @@ import Type from "../type.mjs";
 // Also, in such case add respective call graph edges in Hologram.CallGraph.list_runtime_mfas/1.
 
 const Erlang_Os = {
+  // Start system_time/0
+  // Returns the current system time in nanoseconds.
```
bartblast, line 11: Could you change it to:  "Returns the current system time in native time unit (nanoseconds)."

assets/js/erlang/os.mjs
```diff
@@ -7,6 +7,15 @@ import Type from "../type.mjs";
...
+  // Uses Date.now() which returns milliseconds since Unix epoch.
```
bartblast, line 12: Could you move this comment to the body of the function and change it to: "Date.now() returns milliseconds since Unix epoch."

test/elixir/hologram/ex_js_consistency/erlang/os_test.exs
```diff
@@ -9,6 +9,26 @@ defmodule Hologram.ExJsConsistency.Erlang.OsTest do
...
+      before_ms = System.os_time(:millisecond)
+      result = :os.system_time()
+      after_ms = System.os_time(:millisecond)
+
+      assert is_integer(result)
+      assert result >= before_ms * 1_000_000
+      assert result <= (after_ms + 1) * 1_000_000
+    end
+
+    test "returns different values on subsequent calls" do
```
bartblast: This test should be removed. Since the Erlang documentation explicitly states "This time is **not a monotonically increasing time**", there's no guarantee that:
1. Subsequent calls return different values (could be same within millisecond precision)
2. Values increase over time (clock can jump backwards due to NTP adjustments)

Testing for "different values" is testing behavior that isn't guaranteed by the function spec.

(Please remember about the JS test as well)

test/javascript/erlang/os_test.mjs
```diff
@@ -15,6 +15,51 @@ defineGlobalErlangAndElixirModules();
...
+      const beforeMs = Date.now();
+      const result = systemTime();
+      const afterMs = Date.now();
+
+      assert.ok(Type.isInteger(result));
+      assert.ok(result.value >= BigInt(beforeMs) * 1_000_000n);
+      assert.ok(result.value <= BigInt(afterMs + 1) * 1_000_000n);
+    });
+
+    it("returns positive values", () => {
```
bartblast: This test is redundant. The first test "returns current system time in nanoseconds" already verifies that the value is >= a known positive timestamp (`beforeMs * 1_000_000n`), which implicitly confirms the value is positive.

test/javascript/erlang/os_test.mjs
```diff
@@ -15,6 +15,51 @@ defineGlobalErlangAndElixirModules();
...
+      const result = systemTime();
+
+      assert.ok(result.value > 0n);
+    });
+
+    it("converts correctly to milliseconds", () => {
```
bartblast: This test overlaps significantly with "returns current system time in nanoseconds". Both tests:
- Call the function
- Compare the result against `Date.now()`
- Allow a small tolerance for test execution time

The first test already validates the nanosecond value falls within the expected millisecond range, which implicitly proves correct conversion.

test/javascript/erlang/os_test.mjs
```diff
@@ -15,6 +15,51 @@ defineGlobalErlangAndElixirModules();
 // Always update both together.
 
 describe("Erlang_Os", () => {
+  describe("system_time/0", () => {
```
bartblast, line 18: Please make the Elixir and JS tests consistent.

test/javascript/erlang/os_test.mjs
```diff
@@ -15,6 +15,51 @@ defineGlobalErlangAndElixirModules();
...
+    const systemTime = Erlang_Os["system_time/0"];
+
+    it("returns current system time in nanoseconds", () => {
+      const beforeMs = Date.now();
+      const result = systemTime();
+      const afterMs = Date.now();
+
+      assert.ok(Type.isInteger(result));
```
bartblast: Please use isTrue, isAtLeast, etc.

mward-sudo:
FYI, I've not been able to used the graphite stacked diffs for this work as I suggested I would elsewhere. Damn thing only pushed PR's to my fork and I couldn't figure out how to change that.

bartblast:
Looks good! Thank you @mward-sudo! 👍

PR #641, opened 2026-01-30
Fix URL in interpreter error message

### LINKED ISSUE #640: Broken URL in interpreter error message
I was playing around with Hologram today and ran across one of the functions that is not yet ported to JS. I understand that issue and I might jump in and try to address it. However, I wanted to separately report an issue with the error message itself: "Uncaught HologramInterpreterError: Function :erlang.error/3 is not yet ported. See what to do here: **https://www.hologram.page/TODO**"

It doesn't seem like www.hologram.page resolves in DNS for me, so I can't get to that page. hologram.page without the www works fine.

Should we update the error message? Or add www as an alias? Or both?

I'd be happy to submit a PR for the error message change if that saves you time.

prehnRA:
Closes #640

This PR updates the error message for unported functions with the correct current URL. I'm open to further changes, so let me know if you have any feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages with clearer guidance and updated resource links to check implementation status and request support when needed.
* **Tests**
  * Updated test expectations to match the revised, multi-line error messages.

### DIFF
--- a/assets/js/interpreter.mjs
+++ b/assets/js/interpreter.mjs
@@ -442,7 +442,7 @@ export default class Interpreter {
     globalThis[moduleJsName][`${functionName}/${arity}`] = () => {
       const message =
         `Function :${moduleExName}.${functionName}/${arity} is not yet ported.\n` +
-        `  * Check implementation status: https://hologram.page/reference/client-runtime` +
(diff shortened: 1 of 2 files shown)


bartblast:
Thank you @prehnRA! 💜

PR #642, opened 2026-01-30
Fix case where match placeholder is on right

prehnRA:
I came across this code that caused an unexpected match error once transpiled:

```elixir
  def put_markdown(%MDEx.Document{} = document, markdown, :top = _position) when is_binary(markdown) or is_list(markdown) do
    %{document | buffer: document.buffer ++ List.wrap(markdown)}
  end
```

It seems that the issue comes from the placeholder `_position` being the right side of the match instead of the left. Since that's valid Elixir, I think we should add a case for it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced pattern matching logic to handle placeholder behavior consistently.

* **Tests**
  * Added test coverage for pattern matching edge cases.

### DIFF
--- a/assets/js/interpreter.mjs
+++ b/assets/js/interpreter.mjs
@@ -660,6 +660,10 @@ export default class Interpreter {
       return right;
     }
 
+    if (Type.isMatchPlaceholder(right)) {
+      return left;
+    }
+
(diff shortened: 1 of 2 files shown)


bartblast:
Good catch! Thank you @prehnRA 🚀

PR #652, opened 2026-01-31
Add fallback mechanism for JS formatter

### LINKED ISSUE #639: Biome binary fails on NixOS
**Describe the bug**
It appears that the biome CLI tool that hologram uses is a dynamically linked executable. This means that this is unable to run as-is on my NixOS machine. It would be great if there was a way that we could specify the location of biome instead of requiring the use of the one that you install for us. Similar to the way that the `tailwind` library allows us to specify the `:path` to the version I have installed on my machine. Without this, I am unable to use hologram at all. When I attempt to start my application after having gone through the hologram installation process, I am met with the following

```
** (RuntimeError) Biome formatter failed (probably there were JavaScript syntax errors)
    (hologram 0.6.6) lib/hologram/compiler.ex:361: Hologram.Compiler.format_files/2
    (hologram 0.6.6) lib/mix/tasks/compile/hologram.ex:108: Mix.Tasks.Compile.Hologram.compile/1
    (hologram 0.6.6) lib/mix/tasks/compile/hologram.ex:222: Mix.Tasks.Compile.Hologram.with_lock/2
    (mix 1.19.4) lib/mix/task.ex:499: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.19.4) lib/mix/task.compiler.ex:299: Mix.Task.Compiler.run_compiler/2
    (mix 1.19.4) lib/mix/task.compiler.ex:287: Mix.Task.Compiler.run/4
    (mix 1.19.4) lib/mix/tasks/compile.all.ex:75: Mix.Tasks.Compile.All.do_run/2
    (mix 1.19.4) lib/mix/sync/lock.ex:122: Mix.Sync.Lock.with_lock/3
```

Note that `esbuild` is statically linked, so this is not an issue.

**Reproduction Repository**
Please provide a link to a minimal reproduction repository forked from Hologram Skeleton: https://github.com/bartblast/hologram_skeleton
Repository URL:
I am not linking a repository as this is very specific to NixOS and not something most people will encounter. It would also happen with any project on NixOS.

**To Reproduce**

1. Install NixOS (simple enough if you can run a VM)
2. Setup Elixir/Erlang and a new Phoenix project
3. Install Hologram
4. Attempt to run the application

**Expected behavior**
I would like to see the application boot and run

**Screenshots**

Image

**Environment information**

- Hologram version: 0.6.6
- Elixir version: 1.19.4
- Erlang/OTP version: 28
- Node.js version: 22.21.1
- Operating System: NixOS
- Browser (if applicable):

**Additional context**
I would be more than happy to submit a PR that would allow this type of configuration.

ankhers:
Some platforms disallow execution of dynamically linked binaries. To support these environments, we now:

1. Attempt to run the bundled executable
2. Fall back to the system PATH executable if the bundled version fails
3. Raise an error prompting the user to install via their package manager if both fail

This ensures the formatter works across a wider range of platforms while providing clear guidance when manual installation is required.

resolves #639 

## Summary by CodeRabbit

* **Bug Fixes**
  * Compilation now detects missing or invalid formatter tools earlier and shows a clear, actionable error during compile, preventing later failures.

* **Chores**
  * Added an early step after installing JS dependencies to resolve and validate the formatter binary path, improving reliability of formatter configuration and failing fast if the formatter is unavailable.

### DIFF
--- a/lib/mix/tasks/compile/hologram.ex
+++ b/lib/mix/tasks/compile/hologram.ex
@@ -56,6 +56,8 @@ defmodule Mix.Tasks.Compile.Hologram do
 
     Compiler.maybe_install_js_deps(assets_dir, build_dir)
 
+    opts = maybe_adjust_formatter_bin_path(opts)
+
     {old_module_digest_plt, module_digest_plt_dump_path} =
       Compiler.maybe_load_module_digest_plt(build_dir)
 
@@ -133,6 +135,46 @@ defmodule Mix.Tasks.Compile.Hologram do
     Logger.info("Hologram: compiler finished")
 
     :ok
+  end
+
+  defp maybe_adjust_formatter_bin_path(opts) do
+    system_formatter_path = "biome"
+
+    formatter_bin_path =
+      Enum.find([opts[:formatter_bin_path], system_formatter_path], &bin_available?/1)
+
+    case formatter_bin_path do
+      nil ->
+        raise RuntimeError,
+          message: """
+          Biome Formatter failed to run
+
+          Neither the bundled biome binary nor a system-installed biome could be executed.
+          This can happen on systems where dynamically linked binaries are not supported,
+          such as NixOS or musl-based distributions (e.g., Alpine Linux).
+
+          To fix this, install biome and ensure it's available in your PATH.
+          For installation options, see: https://biomejs.dev/guides/manual-installation/
+          """
+
+      path ->
+        Keyword.put(opts, :formatter_bin_path, path)
+    end
+  end
+
+  defp bin_available?(path) do
+    cmd_args = ["--version"]
+
+    cmd_opts = [parallelism: true, stderr_to_stdout: true]
+
+    try do
+      case SystemUtils.cmd_cross_platform(path, cmd_args, cmd_opts) do
+        {_exit_msg, 0} -> true
+        _cmd_result -> false
+      end
+    rescue
+      _e -> false
+    end
   end
 
   defp build_default_opts do


ankhers:
This is a first pass. I am having some issues running the tests locally because of how NixOS makes certain things available (e.g., you are testing specifically the value `~c"erl"` and on my machine it is `~c".erl-wrapped"`). There are also quite a few numbers in tests that are off by one. Not sure if these are unique to my environment or if they are issues that are in the codebase as these fail on current `dev` for me.

As for the actual implementation, please let me know if you would like to see anything adjusted. I initially tried checking the value inside the `build_default_opts/0` function only to realize that the opts are being created before installing the deps. So this seemed the most sensible way to me. But please let me know if there is something better.

lib/mix/tasks/compile/hologram.ex
```diff
@@ -135,6 +137,32 @@ defmodule Mix.Tasks.Compile.Hologram do
...
+  defp bin_available?(path) do
+    cmd_args = ["--help"]
+
+    cmd_opts = [parallelism: true, stderr_to_stdout: true]
+
+    case SystemUtils.cmd_cross_platform(path, cmd_args, cmd_opts) do
+      {_exist_msg, 0} -> true
+      _ -> false
+    end
+  end
```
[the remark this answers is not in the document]
ankhers: :sweat_smile: This is something I hit in development and forgot to fix before pushing up.

lib/mix/tasks/compile/hologram.ex
```diff
@@ -135,6 +137,37 @@ defmodule Mix.Tasks.Compile.Hologram do
...
+  defp maybe_adjust_formatter_bin_path(opts) do
+    system_formatter_path = "biome"
+
+    formatter_bin_path =
+      Enum.find([opts[:formatter_bin_path], system_formatter_path], &bin_available?/1)
+
+    case formatter_bin_path do
+      nil ->
+        raise RuntimeError,
+          message: "Biome failed to run. Please install biome via your system's package manager."
```
bartblast: Could you expand the error message? Something like:

```elixir
"""
Biome formatter failed to run.

Neither the bundled biome binary nor a system-installed biome could be executed.
This can happen on systems where dynamically linked binaries are not supported,
such as NixOS or musl-based distributions (e.g., Alpine Linux).

To fix this, install biome via your system's package manager and ensure it's
available in your PATH:
  - NixOS: nix-env -iA nixpkgs.biome
  - Alpine: apk add biome
  - Other: https://biomejs.dev/guides/getting-started/#installation
"""
```

lib/mix/tasks/compile/hologram.ex
```diff
@@ -135,6 +137,37 @@ defmodule Mix.Tasks.Compile.Hologram do
...
+  end
+
+  defp bin_available?(path) do
+    cmd_args = ["--help"]
+
+    cmd_opts = [parallelism: true, stderr_to_stdout: true]
+
+    try do
+      case SystemUtils.cmd_cross_platform(path, cmd_args, cmd_opts) do
+        {_exist_msg, 0} -> true
```
bartblast: This variable holds the command's stdout, not an "existence message". Consider renaming to `_output` for clarity.

lib/mix/tasks/compile/hologram.ex
```diff
@@ -135,6 +137,37 @@ defmodule Mix.Tasks.Compile.Hologram do
...
+        raise RuntimeError,
+          message: "Biome failed to run. Please install biome via your system's package manager."
+
+      path ->
+        Keyword.put(opts, :formatter_bin_path, path)
+    end
+  end
+
+  defp bin_available?(path) do
+    cmd_args = ["--help"]
```
bartblast: Nit: Any particular reason for using `--help` over `--version`? Just curious - both work fine, though `--version` produces smaller output.

lib/mix/tasks/compile/hologram.ex
ankhers: I am reluctant to add install instructions, specifically for NixOS. While you certainly CAN run that command and it would install the package for you, people generally do not install packages using that method. It is much more common to have a configuration file somewhere that they maintain in order to install and configure their systems. However, if we have that command there, people that use NixOS would more than likely install it via their preferred method, even if that happens to be via `nix-env`.

Overall, I definitely agree that there needs to be a better message.

lib/mix/tasks/compile/hologram.ex
ankhers: Sadly my fingers betrayed me. This is actually a typo and was meant to be `exit_msg`. Will certainly fix this up.

lib/mix/tasks/compile/hologram.ex
ankhers: It is generally the first command that I run on a new binary. I can switch this to `--version`.

lib/mix/tasks/compile/hologram.ex
```diff
@@ -135,6 +137,49 @@ defmodule Mix.Tasks.Compile.Hologram do
...
+          To fix this, install biome via your systems's package manager and ensure it's
+          available in your PATH
+            - NixOS: nix-env -iA nixpkgs.biome
+            - Alpine: apk add biome
+            - Other: https://biomejs.dev/guides/getting-started/#installation
+          """
+
+      path ->
+        Keyword.put(opts, :formatter_bin_path, path)
+    end
```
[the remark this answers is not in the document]
ankhers: This is a safety check to ensure that a formatter is available instead of breaking further down the chain when we actually attempt to use it.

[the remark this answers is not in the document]
ankhers: That typo has been fixed in 67afd27

lib/mix/tasks/compile/hologram.ex
bartblast: @ankhers I agree that the error message should feel natural and actionable for users of each platform (NixOS, Alpine, etc.). The first part is fine:

> Biome formatter failed to run.
>
> Neither the bundled biome binary nor a system-installed biome could be executed.
> This can happen on systems where dynamically linked binaries are not supported,
> such as NixOS or musl-based distributions (e.g., Alpine Linux).

What would you propose for the message that follows?

ankhers: I think this is something that is hard to do in a single message like this. Take a look at the [install instructions](https://devenv.sh/getting-started/#2-install-devenv) for a tool called devenv. They list 4 different ways to install the `devenv` package. 

I think if people are using nix, and they come across this message as it currently is, I think people will know how to translate the `nix-env` command into whatever their preferred install method would be.

bartblast: I understand, and I agree we don't need distro-specific commands. But what do you suggest for the second part of the message? At minimum we need to tell the user to install biome and ideally link to the docs, e.g.:

```
To fix this, install biome and ensure it's available in your PATH.
For installation options, see: https://biomejs.dev/guides/manual-installation/
```

ankhers: I think this probably makes the most sense. I will make that change.

bartblast:
Thank you @ankhers! 💜

PR #693, opened 2026-02-11
Support custom HTML element tags in template syntax

### LINKED ISSUE #691: Support custom HTML element tags in template syntax
**Describe the bug**
I am attempting to include the `@tailwindplus/elements` package. This will give me a handful of custom HTML elements that I can use to build my application. At least some of the custom elements have the `-` character in the name. e.g., `<el-dialog>`. Full error below

```
== Compilation error in file lib/my_app_web/main_layout.ex ==
** (Hologram.TemplateSyntaxError)

Reason:
Unknown reason.

Hint:
Please report that you received this message here: https://github.com/bartblast/hologram/issues
and include a markup snippet that will allow us to reproduce the issue.

ss="h-full">\n    <el-dialog></el-dialog>\n
                     ^

status = :start_tag

token = {:symbol, "-"}

context = %Hologram.Template.Parser.Context{attribute_name: nil, attribute_value: [], attributes: [], block_name: nil, delimiter_stack: [], node_type: :tag, prev_status: :start_tag_name, processed_tags: [{:text, "\n    "}, {:start_tag, {"body", [{"class", [text: "h-full"]}]}}, {:text, "\n  "}, {:end_tag, "head"}, {:text, "\n  "}, {:self_closing_tag, {"Runtime", []}}, {:text, "\n    "}, {:end_tag, "title"}, {:text, "My App"}, {:start_tag, {"title", []}}, {:text, "\n    "}, {:self_closing_tag, {"meta", [{"name", [text: "csrf-token"]}, {"content", [expression: "{Phoenix.Controller.get_csrf_token()}"]}]}}, {:text, "\n    "}, {:self_closing_tag, {"meta", [{"name", [text: "viewport"]}, {"content", [text: "width=device-width, initial-scale=1"]}]}}, {:text, "\n    "}, {:self_closing_tag, {"meta", [{"charset", [text: "utf-8"]}]}}, {:text, "\n    "}, {:start_tag, {"head", []}}, ...], ...}

    (hologram 0.6.6) lib/hologram/template/parser.ex:1064: Hologram.Template.Parser.raise_error/5
    (hologram 0.6.6) lib/hologram/template.ex:22: Hologram.Template.dom_ast/1
    (hologram 0.6.6) lib/hologram/template.ex:40: Hologram.Template.build_holo_sigil_ast/1
    (hologram 0.6.6) expanding macro: Hologram.Template.sigil_HOLO/2
    (my_app 0.1.0) lib/my_app_web/main_layout.ex:7: MyAppWeb.MainLayout.template/0
```

**Reproduction Repository**
Please provide a link to a minimal reproduction repository forked from Hologram Skeleton: https://github.com/bartblast/hologram_skeleton
Repository URL:
Not including a repo as this is not specific to my setup. I can create one if need be.

**To Reproduce**

1. Include a `-` character in any HTML element

**Expected behavior**
https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#using_a_custom_element -- This should be valid.

**Screenshots**
N/A

**Environment information**

- Hologram version: 0.6.6
- Elixir version: 1.19.5
- Erlang/OTP version: 28
- Node.js version: 24.13.0
- Operating System: NixOS 25.11
- Browser (if applicable):

**Additional context**
N/A

ankhers:
Standard HTML elements do not use hyphens in their names, but custom elements (Web Components) require them, e.g., <my-widget></my-widget>.

Resolves #691 

## Summary by CodeRabbit

* **New Features**
  * Templates now support hyphenated element tag names in start, end, and self-closing forms (e.g., <foo-bar>, </foo-bar>, <foo-bar />), including multiple hyphens and lone-hyphen segments; element names may not start with a hyphen.

* **Tests**
  * Added tests covering parsing of hyphenated tags: multiple hyphens, lone hyphens, attributes, self-closing variants, and matching end tags.

### DIFF
--- a/lib/hologram/template/parser.ex
+++ b/lib/hologram/template/parser.ex
@@ -103,6 +103,9 @@ defmodule Hologram.Template.Parser do
           }
   end
 
+  defguardp is_valid_element_name_token(type, value)
+            when type == :string or (type == :symbol and value == "-")
+
   @doc """
   Parses markup into tags.
 
@@ -248,8 +251,37 @@ defmodule Hologram.Template.Parser do
   end
 
   # --- END TAG NAME ---
+  def parse_tokens(%{tag_name: ""} = context, :end_tag_name, [{:symbol, "-"} = token | rest]) do
+    details =
+      StringUtils.normalize_newlines("""
+      Reason:
+      Element names beginning with '-' are invalid.
 
-  def parse_tokens(context, :end_tag_name, [{:string, tag_name} = token | rest]) do
+      Hint:
+      Prefix your WebComponent name with at least one lower case letter.
+      """)
+
+    raise_error(details, context, :end_tag_name, token, rest)
+  end
+
+  def parse_tokens(
+        context,
+        :end_tag_name,
+        [{first_type, first_value} = token, {second_type, second_value} | _rest] = tokens
+      )
+      when is_valid_element_name_token(first_type, first_value) and
+             is_valid_element_name_token(second_type, second_value) do
+    context
+    |> set_tag_name(context.tag_name <> first_value)
+    |> add_processed_token(token)
+    |> parse_tokens(:end_tag_name, tl(tokens))
+  end
+
+  # Anything else (or no more tokens) -> name is complete, transition
+  def parse_tokens(context, :end_tag_name, [{type, value} = token | rest])
+      when type == :string or (type == :symbol and value == "-") do
+    tag_name = context.tag_name <> value
+
     context
     |> set_tag_name(tag_name)
     |> maybe_disable_script_mode(tag_name)
@@ -430,8 +462,37 @@ defmodule Hologram.Template.Parser do
   end
 
   # --- START TAG NAME ---
+  def parse_tokens(%{tag_name: ""} = context, :start_tag_name, [{:symbol, "-"} = token | rest]) do
+    details =
+      StringUtils.normalize_newlines("""
+      Reason:
+      Element names beginning with '-' are invalid.
 
-  def parse_tokens(context, :start_tag_name, [{:string, tag_name} = token | rest]) do
+      Hint:
+      Prefix your WebComponent name with at least one lower case letter.
+      """)
+
+    raise_error(details, context, :start_tag_name, token, rest)
+  end
+
+  def parse_tokens(
+        context,
+        :start_tag_name,
+        [{first_type, first_value} = token, {second_type, second_value} | _rest] = tokens
+      )
+      when is_valid_element_name_token(first_type, first_value) and
+             is_valid_element_name_token(second_type, second_value) do
+    context
+    |> set_tag_name(context.tag_name <> first_value)
+    |> add_processed_token(token)
+    |> parse_tokens(:start_tag_name, tl(tokens))
+  end
+
+  # Anything else (or no more tokens) -> name is complete, transition
+  def parse_tokens(context, :start_tag_name, [{type, value} = token | rest])
+      when type == :string or (type == :symbol and value == "-") do
+    tag_name = context.tag_name <> value
+
     context
     |> reset_attributes()
     |> set_tag_name(tag_name)
@@ -748,18 +809,21 @@ defmodule Hologram.Template.Parser do
     |> reset_token_buffer()
     |> add_processed_token(token)
     |> set_prev_status(:text)
+    |> set_tag_name("")
     |> parse_tokens(:end_tag_name, rest)
   end
 
   def parse_tokens(%{node_type: :text, script?: false} = context, :text, [
-        {:symbol, "<"} = token | [{:string, _value} | _tokens] = rest
-      ]) do
+        {:symbol, "<"} = token | [{next_type, next_value} | _tokens] = rest
+      ])
+      when is_valid_element_name_token(next_type, next_value) do
     context
     |> maybe_add_text_tag()
     |> reset_token_buffer()
     |> add_processed_token(token)
     |> set_prev_status(:text)
     |> set_node_type(:tag)
+    |> set_tag_name("")
     |> parse_tokens(:start_tag_name, rest)
   end
 
--- a/test/elixir/hologram/template/parser_test.exs
+++ b/test/elixir/hologram/template/parser_test.exs
@@ -71,7 +71,7 @@ defmodule Hologram.Template.ParserTest do
     end
   end
 
-  describe "element tags" do
+  describe "standard element tags" do
     test "non-void HTML element start tag" do
       assert parse_markup("<div>") == [start_tag: {"div", []}]
     end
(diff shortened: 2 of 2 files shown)


ankhers:
Good catch on the allowed names. Though, as far as I can tell they are unable to start with a `-`. So `<-foo>` would be invalid. I have implemented tests and fixed up the parser to include the rest of the functionality.

ankhers:
Let me know if you want me to implement the test suggested by coderabbit. I'm not sure if it would technically be a duplicate or not based on how the parser works.

bartblast:
The first one is beneficial IMO, since we're parsing two tokens at once here and that's a separate code path. The second one not so much, kind of redundant.

ankhers:
Implemented the first test.

lib/hologram/template/parser.ex
```diff
@@ -483,6 +500,27 @@ defmodule Hologram.Template.Parser do
...
+
+  def parse_tokens(%{prev_status: :start_tag_name} = context, :start_tag, [
+        {:symbol, "-"} = token | rest
+      ]) do
+    context
+    |> set_tag_name(context.tag_name <> "-")
+    |> add_processed_token(token)
+    |> set_prev_status(:start_tag_name)
+    |> parse_tokens(:start_tag, rest)
+  end
```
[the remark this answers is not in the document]
ankhers: Should be resolved in aa31dd2. I chose to instead maybe enable script mode only when the closing tag (`/>` or `>`) is found.

lib/hologram/template/parser.ex
```diff
@@ -476,11 +492,36 @@ defmodule Hologram.Template.Parser do
   end
 
   def parse_tokens(context, :start_tag, [{:symbol, "/>"} = token | rest]) do
-    parse_start_tag_end(context, token, rest, true)
+    context
+    |> maybe_enable_script_mode(context.tag_name)
```
bartblast: Isn't `maybe_enable_script_mode/2` call redundant for self-closed cases?

lib/hologram/template/parser.ex
```diff
@@ -476,11 +492,36 @@ defmodule Hologram.Template.Parser do
...
   end
 
   def parse_tokens(context, :start_tag, [{:symbol, ">"} = token | rest]) do
-    parse_start_tag_end(context, token, rest, false)
+    context
+    |> maybe_enable_script_mode(context.tag_name)
+    |> parse_start_tag_end(token, rest, false)
+  end
+
+  def parse_tokens(%{prev_status: :start_tag_name} = context, :start_tag, [
```
bartblast: Could you move this 2 to `# --- START TAG NAME ---` section?

test/elixir/hologram/template/parser_test.exs
```diff
@@ -111,6 +111,66 @@ defmodule Hologram.Template.ParserTest do
...
+    test "multi-hyphenated element start tag" do
+      assert parse_markup("<foo-bar-baz>") == [start_tag: {"foo-bar-baz", []}]
+    end
+
+    test "hyphenated ending element start tag" do
+      assert parse_markup("<foo-bar->") == [start_tag: {"foo-bar-", []}]
+    end
+
+    test "element name cannot start with a hyphen" do
+      assert_raise TemplateSyntaxError, fn ->
```
bartblast: Could you test the error message?

lib/hologram/template/parser.ex
```diff
@@ -435,7 +452,6 @@ defmodule Hologram.Template.Parser do
     context
     |> reset_attributes()
     |> set_tag_name(tag_name)
-    |> maybe_enable_script_mode(tag_name)
```
bartblast, line 438: Moving `maybe_enable_script_mode` from `:start_tag_name` to the `:start_tag` `">"` / `"/>"` handlers misses another path that closes start tags:

```elixir
def parse_tokens(context, :attribute_name, [{:symbol, ">"} = token | rest]) do
...
```

`<script defer>` triggers this path (boolean attribute immediately before `>`), so script mode is never enabled. `<script defer >` (with a space) still works because whitespace returns to `:start_tag` first.

How would a solution look that keeps setting script mode in `:start_tag_name` (the original placement)? That way there's a single place that handles it, instead of needing to patch every clause that can close a start tag - which is fragile if more constructs are introduced in the future.

test/elixir/hologram/template/parser_test.exs
```diff
@@ -111,6 +111,66 @@ defmodule Hologram.Template.ParserTest do
...
+    end
+
+    test "hyphenated ending element end tag" do
+      assert parse_markup("</foo-bar->") == [end_tag: "foo-bar-"]
+    end
+
+    test "multiple simultaneous hyphenated element end tag" do
+      assert parse_markup("</foo--bar>") == [end_tag: "foo--bar"]
+    end
   end
```
bartblast, line 230: Since there are quite a few hyphenated tag tests now, it might be worth extracting them into a separate `describe "custom element tags"` block (and renaming the current one to `"standard element tags"`). This would keep things organized as more cases are added.

Also a few tests I think are missing:
- Leading hyphen in end tags (`</-foo>`) should raise
- Round-trip: `<foo-bar>text</foo-bar>`
- Script mode is NOT enabled for `<script-widget>...</script-widget>`
- Script mode IS still enabled for `<script>` with attributes, e.g. `<script defer>var x = "<div>";</script>`
- etc.

lib/hologram/template/parser.ex
ankhers: You are absolutely right.

lib/hologram/template/parser.ex
ankhers: I am thinking something like this.

```elixir
  def parse_tokens(context, :start_tag_name, [{:symbol, "-"} = token | rest]) do
    context
    |> set_tag_name(context.tag_name <> "-")
    |> add_processed_token(token)
    |> parse_tokens(:start_tag_name, rest)
  end

  def parse_tokens(context, :start_tag_name, [{:string, tag_name} = token | rest]) do
    context
    |> set_tag_name(context.tag_name <> tag_name)
    |> add_processed_token(token)
    |> parse_tokens(:start_tag_name, rest)
  end

  def parse_tokens(context, :start_tag_name, tokens) do
    context
    |> reset_attributes()
    |> maybe_enable_script_mode(context.tag_name)
    |> set_prev_status(:start_tag_name)
    |> parse_tokens(:start_tag, tokens)
  end
```

This will require a slight change to the case when matching on `<` in order to start parsing the `:start_tag_name`. We would need to `set_tag_name` to `""`. There may be a better place to do this as well.

The reason for this is that the original implementation (current dev) expected the `context.tag_name` to have been overwritten by the single `:string` that makes up the name. With this change, multiple `:string`s can create the tag name, meaning we potentially need to concatenate multiple values to create the tag name. The fact that `context.tag_name` is never cleared/removed meant that something like 

```elixir
Hologram.Template.Parser.parse_markup("<script></script><script></script>")
```

would end up being parsed as

```elixir
[
  start_tag: {"script", []},
  end_tag: "script",
  start_tag: {"scriptscript", []},
  end_tag: "scriptscript"
]
```

Notice how the second set of tags is `scriptscript` instead of the expected `script`.

The above implementation should allow parsing only the `:string`s and `symbol: "-"` for a tag name. After it reaches another token, such as `:whitespace` or a different `:symbol`, it will check if it should enable script mode and continue parsing as expected.

This does have the downside that there is an extra function call for each `:start_tag_name` because of the third function head that acts as a catch-all because I thought it would be worse to explicitly check for either `:whitespace` to start (potentially) checking for attributes or check for a `symbol: ">"` in order to close the tag. With this catch-all in place parsing will continue as it currently does.

Let me know if this approach is in line with what you were thinking or if there is another approach that you would like me to explore as well.

bartblast: @ankhers Thanks for the detailed write-up! I like the idea of resetting the tag name to avoid the concatenation issue with consecutive tags. However, I'd like to avoid the catch-all and the extra dispatch it introduces for every tag.

What if instead we call `maybe_disable_script_mode` when `{:symbol, "-"}` is encountered in `:start_tag_name`? That would undo script mode if it was prematurely enabled for something like `<script-element>`.

ankhers: I'm not sure that we can do that. The main reason for the catch-all is to move from `:start_tag_name` to `:start_tag`. Otherwise we won't really be able to make that transition. I just also threw the script mode check in there so that there is a single place for it instead of doing it in multiple places.

Previously there was only one valid case for `:start_tag_name`. That being the whole name of the element. This meant that as soon as you have this, you could move on to `:start_tag`. 

Unfortunately now, there are multiple valid cases for `:start_tag_name`. The minimum of a string as the first part of the name, then optionally some number of `-` and other strings. Ultimately it will be one of `>` or `/>` in order to close the tag, or some kind of whitespace which is likely to start the attributes. Though it could also be whitespace and then closing the tag as well.

If we wanted to avoid the extra dispatch per tag, we would require three additional function heads, 

1. Check for `>` - which would allow us to skip checking for attributes and such, but there is now 2 places where we are checking for this
2. Check for `/>` - the same issue as above
3. Check for whitespace - which could allow us to move on to checking for attributes

I think for complexity purposes, I would prefer the additional function call per element, but I will admit I do not know how much overhead that would add to the process, especially for larger documents.

bartblast: How about a lookahead approach instead of the catch-all? Something along the lines of:

```elixir
# Next token continues the name -> accumulate, stay in :start_tag_name
def parse_tokens(context, :start_tag_name, [{:string, tag_name} = token, {next_type, next_value} | _rest] = tokens)
    when next_type == :string or (next_type == :symbol and next_value == "-") do
  context
  |> set_tag_name(context.tag_name <> tag_name)
  |> add_processed_token(token)
  |> parse_tokens(:start_tag_name, tl(tokens))
end

# Hyphen -> consume it, stay in :start_tag_name
def parse_tokens(context, :start_tag_name, [{:symbol, "-"} = token | rest]) do
  context
  |> set_tag_name(context.tag_name <> "-")
  |> add_processed_token(token)
  |> parse_tokens(:start_tag_name, rest)
end

# Anything else (or no more tokens) -> name is complete, transition
def parse_tokens(context, :start_tag_name, [{:string, tag_name} = token | rest]) do
  tag_name = context.tag_name <> tag_name

  context
  |> reset_attributes()
  |> set_tag_name(tag_name)
  |> maybe_enable_script_mode(tag_name)
  |> add_processed_token(token)
  |> set_prev_status(:start_tag_name)
  |> parse_tokens(:start_tag, rest)
end
```

The first clause peeks at the next token - if it's a string or `"-"`, the name isn't finished yet. Otherwise the last clause knows the name is complete and transitions to `:start_tag`. This avoids the catch-all and the extra dispatch per tag, while keeping `maybe_enable_script_mode` in a single place. We'd still need the `set_tag_name("")` reset on `<` that you mentioned.

ankhers: Great idea. I will look at implementing this.

ankhers: I was not able to do quite exactly what you had here, but I was able to adapt it to what was needed. The big change here is that instead of the `when` clause that is listed, I created a separate guard in order to collapse the first two function clauses into one. This allowed the support of the trailing `-` character. The custom guard also helps because we needed to do the exact same thing for `:end_tag_name`. 

I also slightly modified (using the guard) the case where we transition to `:start_tag_name` so that we can give a proper error message if someone were to accidentally prefix their tag name with the `-`. 

Please let me know if there are additional changes that you would like to see happen. Please double check the tests specifically to ensure that I covered everything that you wanted to see.

bartblast:
Hey @ankhers, I noticed you pushed some updates - is this ready for another review?

ankhers:
It should be, yes.

ankhers:
Hey @bartblast, just wondering if you have had any time to take a look at these changes? No rush if you are busy.

bartblast:
Hey! Thanks for the patience. I'm currently finishing up work on JS interop - once that's wrapped up I'll review this. Shouldn't be too long!

bartblast:
Looks good, thank you @ankhers! 💜

ankhers:
Thanks for working through this with me! If you have any tickets that you would like to send my way, please do. I would love to contribute more towards this project!

bartblast:
Hey @ankhers, sorry for the late reply - I must have missed your comment somehow!

The best way you could help right now is with bug fixes - that's tremendous help for me because it allows me to focus on design work, which is needed to move a few subprojects forward.

I created an ElixirForum thread with tasks for contributors: https://elixirforum.com/t/contributor-friendly-tasks-to-claim/74339

Feel free to pick up anything that interests you! If you choose something, just leave a comment on the issue so I can assign it to you - that way others know someone is already working on it.

I'll try to keep adding more tasks for contributors in that thread over time.

PR #695, opened 2026-02-13
Fix error message resolution for error types without a message field

### LINKED ISSUE #694: Error message resolution crashes for error types without a message field

bartblast:
Closes #694

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error message resolution with enhanced fallback handling for edge cases where message fields may be unavailable.

* **Refactor**
  * Refactored internal error message extraction mechanism to use struct-based resolution instead of wrapper-based approach.

### DIFF
--- a/assets/js/errors/boxed_error.mjs
+++ b/assets/js/errors/boxed_error.mjs
@@ -10,7 +10,7 @@ export default class HologramBoxedError extends Error {
     this.struct = struct;
 
     const boxedType = Interpreter.getErrorType(this);
-    const boxedMessage = Interpreter.resolveErrorMessage(this.struct);
+    const boxedMessage = Interpreter.resolveErrorMessage(struct);
 
     this.message = `(${boxedType}) ${boxedMessage}`;
   }


bartblast:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > test/javascript/support/helpers.mjs (1)> `76-86`: _⚠️ Potential issue_ | _🟡 Minor_
> > > **Align non-regex comparisons with resolved messages.**
> > > Right now, non-regex assertions still compare against `Type.errorStruct(...)`, which will fail for error structs without a `message` field even when the resolved message matches. Consider comparing the resolved message string for non-regex expectations.
> > > 
> > > 🛠️ Proposed fix
> > > ```diff
> > > -  const messageMatches = isRegex
> > > -    ? expectedErrorMessage.test(receivedErrorMessage)
> > > -    : Interpreter.isStrictlyEqual(
> > > -        error.struct,
> > > -        Type.errorStruct(expectedErrorType, expectedErrorMessage),
> > > -      );
> > > +  const messageMatches = isRegex
> > > +    ? expectedErrorMessage.test(receivedErrorMessage)
> > > +    : receivedErrorMessage === expectedErrorMessage;
> > > ```

@coderabbitai The non-regex branch intentionally does a full structural comparison rather than just comparing the message string. This catches subtle bugs that a plain string comparison would miss. All non-regex test assertions use standard error types that have a message field, so this works correctly in practice.

PR #704, opened 2026-02-16
Extracts UTF-8 prefix validation from :unicode module to BitString utility class

### LINKED ISSUE #703: Refactor :unicode module - extract UTF-8 prefix validation to BitString
Extract duplicated UTF-8 validation logic into a centralized method in the Bitstring class, eliminating code duplication across multiple Unicode normalization functions.

  - Validate UTF-8 byte sequences
  - Finds the longest valid UTF-8 prefix
  - Detect truncated sequences (incomplete but potentially valid UTF-8)
  - Return `{validLength: number, isTruncated: boolean}`
  - Handles edge cases: overlong encodings, surrogates, invalid continuation bytes

mward-sudo:
Closes #703 

## Summary by CodeRabbit

* **New Features**
  * Added a UTF-8 prefix validation utility to detect the longest valid UTF-8 prefix and whether trailing bytes are an incomplete sequence.

* **Refactor**
  * Centralized UTF-8 validation across Unicode transformation flows for more consistent handling and normalization of invalid or incomplete inputs.

* **Tests**
  * Expanded test coverage with extensive valid, invalid, truncated, and edge-case UTF-8 scenarios.

### DIFF
--- a/assets/js/bitstring.mjs
+++ b/assets/js/bitstring.mjs
@@ -793,15 +793,6 @@ export default class Bitstring {
 
   // Finds the length of the longest valid UTF-8 prefix and whether the tail is truncated.
   static utf8PrefixValidation(bitstring) {
-    // Helper: Determines the expected length of a UTF-8 sequence based on the leader byte.
-    const getSequenceLength = (leaderByte) => {
-      if ((leaderByte & 0x80) === 0) return 1;
-      if ((leaderByte & 0xe0) === 0xc0) return 2;
-      if ((leaderByte & 0xf0) === 0xe0) return 3;
-      if ((leaderByte & 0xf8) === 0xf0) return 4;
-      return -1;
-    };
-
     // Helper: Decodes a UTF-8 code point from a sequence of bytes starting at 'start' with the given 'length'.
     const decodeCodePoint = (start, length) => {
       const extractSingleByte = () => data[start];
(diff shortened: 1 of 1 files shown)


bartblast:
@mward-sudo, the direction is good - extracting shared logic to `Bitstring` as we discussed. However, I'd like us to go more granular. Right now `utf8PrefixValidation` bundles 7 helpers as inline closures, which means they can't be reused or unit tested individually. If we add proper unit tests for each of those helpers at this level, we're back to multi-thousand-line PRs. And without extracting them, we'll end up duplicating them again as more encoding work lands.

Also, a naming note: `utf8PrefixValidation` is a noun - I'd prefer a verb-based name like `validateUtf8Prefix` to stay consistent with the rest of the codebase.

Here's the workflow I'd like us to follow, using `getSequenceLength` as an example:

1. **Extract** it as a static method on `Bitstring` (e.g. `Bitstring.getUtf8SequenceLength(leaderByte)`)
2. **Add unit tests** for that method
3. **Replace** inline calls with the new shared method

That's one PR. Then repeat for the next building block. Once the small pieces are in place, composing them into `validateUtf8Prefix` becomes a straightforward PR itself.

This should be simpler for both of us - each PR becomes a small, mechanical task (extract one function, test it, replace calls) with no ambiguity about scope, and I can review and merge them quickly instead of going back and forth on a larger PR.

Could you restructure along these lines?

PR #710, opened 2026-02-17
Extract UTF-8 code point decoding to Bitstring class

### LINKED ISSUE #709: Extract UTF-8 code point decoding to Bitstring class
As `decodeUtf8CodePoint(bytes, start, length)`

mward-sudo:
Closes #709 

## Summary by CodeRabbit

* **New Features**
  * Public UTF‑8 code point decoding exposed for consistent text handling.

* **Improvements**
  * Unified decoding across text processing paths and aligned invalid-sequence handling with each normalization form (NFC/NFD/NFKC/NFKD).
  * Preserved existing error/invalid semantics while centralizing decoding.

* **Tests**
  * Added tests covering 1–4 byte decoding, non-zero offsets, and malformed/truncated sequence cases.

### DIFF
--- a/assets/js/bitstring.mjs
+++ b/assets/js/bitstring.mjs
@@ -247,6 +247,27 @@ export default class Bitstring {
     }
   }
 
+  // Decodes a UTF-8 sequence starting at the given position.
+  // Returns the decoded Unicode code point value.
+  // bytes: Uint8Array containing the UTF-8 encoded data
+  // start: byte index where the sequence begins
+  // length: number of bytes in the UTF-8 sequence (1-4)
+  static decodeUtf8CodePoint(bytes, start, length) {
+    if (length === 1) return bytes[start];
+
+    // First byte masks: 2-byte=0x1f, 3-byte=0x0f, 4-byte=0x07
+    const firstByteMasks = {2: 0x1f, 3: 0x0f, 4: 0x07};
+
+    let codePoint = bytes[start] & firstByteMasks[length];
+
+    // Process continuation bytes (all use 0x3f mask, shift by 6 each)
+    for (let i = 1; i < length; i++) {
+      codePoint = (codePoint << 6) | (bytes[start + i] & 0x3f);
+    }
+
+    return codePoint;
+  }
+
   static fromBits(bits) {
     const bitCount = bits.length;
     const byteCount = Math.ceil(bitCount / 8);
--- a/assets/js/erlang/unicode.mjs
+++ b/assets/js/erlang/unicode.mjs
@@ -98,34 +98,6 @@ const Erlang_Unicode = {
       // Checks if a byte is a valid UTF-8 continuation byte (10xxxxxx).
       const isValidContinuation = (byte) => (byte & 0xc0) === 0x80;
 
-      // Decodes a UTF-8 sequence starting at the given position.
-      // Returns the decoded Unicode code point value.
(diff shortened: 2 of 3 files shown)


mward-sudo:
I'd appreciate some advice on the best way to validate function parameters in this context.

For completeness (as this is a public function in the class) we should validate parameters for correctness, and CodeRabbit has made some suggestions for further validations that need checking.

I'm thinking that I need to maintain the following:

1) Erlang correct errors from the calling module functions
2) Prefer to avoid try/catch blocks at the calling site.
3) Validate the function parameters properly.
4) Still provide proper feedback to developers about the nature of the validation failure.

Due to the validation issues, the code for the PR is quickly becoming larger than I intended.

mward-sudo:
Follow up thoughts:

- The Erlang port code has the responsibility to:
  - Report Erlang style errors
  - Not call the utility function with invalid args
- Therefore, it's acceptable to blow up with an unhandled exception in the utility class as something is wrong with the port code if invalid args are passed to the utility function.

Maybe??

bartblast:
The extracted utilities should be kept simple and trust the caller.

So to directly answer your follow-up: yes, it's fine to let the utility blow up with an unhandled exception if it receives invalid args - that would indicate a bug in the port code, not a user error. The port code is responsible for validating inputs and producing Erlang-style errors; the utility doesn't need to duplicate that.

That said, if a utility does throw, I don't have a problem with using a `try` statement in the ported function to convert it to an Erlang-style error specific to the caller. What we should avoid is using `catch` as a control flow mechanism (i.e. for regular processing logic rather than genuinely exceptional cases).

The reason: `try` blocks are essentially free in modern JS engines - when no exception is thrown, the engine just sets up a handler entry point with virtually zero overhead. The `catch` path, however, is expensive: the engine has to unwind the call stack, generate a stack trace, and - in V8 - the function may get deoptimized by TurboFan, meaning it falls back to slower unoptimized code on subsequent calls. So using `catch` for expected/normal control flow would be a performance concern, but for genuinely exceptional errors it's perfectly fine.

Does that answer your questions?

mward-sudo:
Yes, I'm happy with that. 

I'm going to keep the validations simple, obvious & performant, throw on invalid args. As discussed, because the client to this API is internal then it should only throw on bugged project code.

assets/js/bitstring.mjs
```diff
@@ -247,6 +247,49 @@ export default class Bitstring {
...
+  // Returns the decoded Unicode code point value.
+  // bytes: Uint8Array containing the UTF-8 encoded data
+  // start: byte index where the sequence begins
+  // length: number of bytes in the UTF-8 sequence (1-4)
+  static decodeUtf8CodePoint(bytes, start, length) {
+    // Basic parameter validations only. For performance reasons we assume the caller has already
+    // determined the correct length from the leader byte and that the bytes are well-formed
+    // UTF-8.
+
+    if (length < 1 || length > 4) {
```
bartblast: Do we need the validations below? Shouldn't we trust the caller? See the comment above and https://github.com/bartblast/hologram/pull/710#issuecomment-3921432233.
The utils will be used only by the runtime or ported code, where we have control.

mward-sudo: I'll drop all the validations.

bartblast:
Thank you @mward-sudo! 💜

PR #715, opened 2026-02-18
Extract UTF-8 continuation byte validation to Bitstring class

### LINKED ISSUE #711: Extract UTF-8 continuation byte validation to Bitstring class
As `BitString.isValidUtf8ContinuationByte(byte)`

mward-sudo:
Closes #711

### Dependencies

- [x] #710

_Please note that this PR includes commits from the PR(s) it is dependent upon. Once the dependant PR(s) are merged to dev branch then this PR will be rebased and will then only contain its own commits. This PR will remain in draft until that point._

## Summary by CodeRabbit

* **New Features**
  * Added a public helper to validate UTF‑8 continuation bytes.

* **Refactor**
  * Centralized continuation-byte validation to reuse the new helper across UTF‑8 decoding paths for consistent handling.

* **Tests**
  * Added tests covering valid and invalid UTF‑8 continuation-byte patterns and related decoding edge cases.

### DIFF
--- a/test/javascript/bitstring_test.mjs
+++ b/test/javascript/bitstring_test.mjs
@@ -5304,6 +5304,16 @@ describe("Bitstring", () => {
     });
   });
 
+  describe("isValidUtf8ContinuationByte()", () => {
+    it("valid continuation byte (10xxxxxx pattern)", () => {
+      assert.isTrue(Bitstring.isValidUtf8ContinuationByte(0x80)); // 10000000
+    });
+
+    it("invalid continuation byte (not 10xxxxxx pattern)", () => {
+      assert.isFalse(Bitstring.isValidUtf8ContinuationByte(0x00)); // 00000000
+    });
+  });
+
   describe("maybeResolveHex()", () => {
     it("when hex field is already set", () => {
       const bitstring = Type.bitstring("Hologram");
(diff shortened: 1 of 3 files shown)


mward-sudo:
@coderabbitai Don't do any further reviews on this PR while it is in draft,

mward-sudo:
@coderabbitai resume, full review and update summary

test/javascript/bitstring_test.mjs
```diff
@@ -5304,6 +5304,20 @@ describe("Bitstring", () => {
     });
   });
 
+  describe("isValidUtf8ContinuationByte()", () => {
```
bartblast, line 5307: I think we can simplify to two assertions, one per equivalence class:

```js
assert.isTrue(Bitstring.isValidUtf8ContinuationByte(0x80));
assert.isFalse(Bitstring.isValidUtf8ContinuationByte(0x00));
```

The contract is "does this byte match `10xxxxxx`?" - so there are really only two classes: matches and doesn't. The boundary/multi-value tests work, but they reason about numeric range boundaries rather than the bit pattern the spec defines, which is a level below the contract.

As a general rule of thumb: one representative test per equivalence class keeps things easy to maintain without losing coverage.

bartblast (changes requested):

PR #692, opened 2026-02-10
Implement with expressions

prehnRA:
Addresses #690 

## Summary by CodeRabbit

* **New Features**
  * Fully implemented the "with" control construct end-to-end; compiler and runtime now support clauses, guards, and else-branches.

* **Bug Fixes**
  * Unified clause processing for case/with flows and standardized error reporting for unmatched clauses and guards.
  * Restored correct with-else behavior, context propagation, and shadowing semantics.

* **Tests**
  * Added extensive compiler and interpreter tests for matches, guards, else-clauses, context propagation, shadowing, and error scenarios.

PR #735, opened 2026-03-12
Add IR for with statements

prehnRA:
# Adds IR for `with` syntax

PR 1 of 5 for with

## Summary by CodeRabbit

* **Refactor**
  * Improved compiler internals to represent and handle distinct clause types within conditional constructs, providing more precise structure, stronger typing, and clearer error reporting during compilation.

### DIFF
--- a/lib/hologram/compiler/ir.ex
+++ b/lib/hologram/compiler/ir.ex
@@ -47,6 +47,7 @@ defmodule Hologram.Compiler.IR do
           | IR.TupleType.t()
           | IR.Variable.t()
           | IR.With.t()
+          | IR.WithClause.t()
 
   defmodule AnonymousFunctionCall do
     @moduledoc false
@@ -384,13 +385,24 @@ defmodule Hologram.Compiler.IR do
     @type t :: %__MODULE__{name: atom, version: integer | nil}
   end
 
-  # TODO: finish implementing
   defmodule With do
     @moduledoc false
 
-    defstruct []
+    defstruct [:clauses, :body, :else_clauses]
 
-    @type t :: %__MODULE__{}
+    @type t :: %__MODULE__{
+            clauses: list(IR.WithClause.t()),
+            body: IR.t(),
+            else_clauses: list(IR.Clause.t())
+          }
+  end
+
+  defmodule WithClause do
+    @moduledoc false
+
+    defstruct [:match, :guards, :expression]
+
+    @type t :: %__MODULE__{match: IR.t(), guards: list(IR.t()), expression: IR.t()}
   end
 
   @doc """
--- a/lib/hologram/compiler/ir.ex
+++ b/lib/hologram/compiler/ir.ex
@@ -47,7 +47,8 @@ defmodule Hologram.Compiler.IR do
           | IR.TupleType.t()
           | IR.Variable.t()
           | IR.With.t()
-          | IR.WithClause.t()
+          | IR.WithBareClause.t()
+          | IR.WithMatchClause.t()
 
   defmodule AnonymousFunctionCall do
     @moduledoc false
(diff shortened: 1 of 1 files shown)


prehnRA:
I think the coderabbit comments are just telling me to do the rest of the code that will go in the other PRs

lib/hologram/compiler/ir.ex
```diff
@@ -384,13 +385,24 @@ defmodule Hologram.Compiler.IR do
...
   defmodule With do
     @moduledoc false
 
-    defstruct []
+    defstruct [:clauses, :body, :else_clauses]
 
-    @type t :: %__MODULE__{}
+    @type t :: %__MODULE__{
+            clauses: list(IR.WithClause.t()),
+            body: IR.t(),
```
bartblast: Could you change the `body` type from `IR.t()` to `IR.Block.t()` - the `do` block can contain multiple expressions.

lib/hologram/compiler/ir.ex
```diff
@@ -384,13 +385,24 @@ defmodule Hologram.Compiler.IR do
...
-  # TODO: finish implementing
   defmodule With do
     @moduledoc false
 
-    defstruct []
+    defstruct [:clauses, :body, :else_clauses]
 
-    @type t :: %__MODULE__{}
+    @type t :: %__MODULE__{
+            clauses: list(IR.WithClause.t()),
```
bartblast: After applying the changes from my top-level comment this will be type union: `IR.WithMatchClause.t() | IR.WithBareClause.t()`.

bartblast (changes requested):
Thank you for the PR! I think we need a few updates:

Elixir `with` supports bare expressions (not just `<-` clauses) between clauses:

```elixir
with {:ok, width} <- Map.fetch(opts, :width),
     double_width = width * 2,
     {:ok, height} <- Map.fetch(opts, :height) do
  {:ok, double_width * height}
end
```

(take a look at the docs: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1)

In the AST, these show up as regular expressions in the clause list (e.g. `{:=, [], [var, expr]}`), alongside the `{:<-, [], [pattern, expr]}` clauses. They always succeed and just bind/execute inline.

The current `WithClause` struct models only `<-` clauses. To handle bare expressions too, please add a separate `WithBareClause` struct:

```elixir
defmodule WithBareClause do
  @moduledoc false

  defstruct [:expression]

  @type t :: %__MODULE__{expression: IR.t()}
end
```

Also rename the existing `WithClause` to `WithMatchClause` to make the distinction clear. Then update `With.clauses` to accept both:

```elixir
@type t :: %__MODULE__{
        clauses: list(IR.WithMatchClause.t() | IR.WithBareClause.t()),
        body: IR.Block.t(),
        else_clauses: list(IR.Clause.t())
      }
```

lib/hologram/compiler/ir.ex
```diff
@@ -384,13 +385,24 @@ defmodule Hologram.Compiler.IR do
...
+            else_clauses: list(IR.Clause.t())
+          }
+  end
+
+  defmodule WithClause do
+    @moduledoc false
+
+    defstruct [:match, :guards, :expression]
+
+    @type t :: %__MODULE__{match: IR.t(), guards: list(IR.t()), expression: IR.t()}
```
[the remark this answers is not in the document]
bartblast: This can be ignored.

prehnRA:
I originally modeled the bare clauses as a WithClause with no guards and a match placeholder for match (comment here laying out the options: https://github.com/bartblast/hologram/pull/692#issuecomment-3946482162). The result was that the match part always "matched" with no changes to the context and the expression would run.

But I can see the value of splitting it out as a separate struct, so I made that change.

bartblast:
> I originally modeled the bare clauses as a WithClause with no guards and a match placeholder for match (comment here laying out the options: [#692 (comment)](https://github.com/bartblast/hologram/pull/692#issuecomment-3946482162)). The result was that the match part always "matched" with no changes to the context and the expression would run.

The split into separate structs isn't just cosmetic - it's needed for correct scoping. With a single clause type, the interpreter evaluates the expression first, then matches the pattern, both writing to the same context. For a `<-` clause like `x <- y = 2`, the `y` binding from expression evaluation would incorrectly leak into subsequent clauses. In Elixir, only the left-side pattern bindings are exported from `<-` clauses. Separate structs let the interpreter handle each case with the right scoping semantics - `WithMatchClause` will need to evaluate the expression in an isolated context and only propagate bindings from the pattern match.

Looks great, merging - thank you! 💜

PR #736, opened 2026-03-12
Implement 'with' expression transformer

prehnRA:
# Add transformer for `with` syntax

- See https://github.com/bartblast/hologram/pull/735 first

PR 2 of 5 for with syntax

## Release Notes

* **New Features**
  * Added complete support for `with` expressions, enabling pattern matching with optional guards, error handling through else clauses, and proper handling of multiple sequential expressions.

* **Tests**
  * Introduced comprehensive test coverage for `with` expressions including pattern matching, guard conditions, else handlers, and multiple fixture modules covering various real-world usage patterns.

* **Chores**
  * Updated code quality tooling configuration.


bartblast:
@prehnRA Now that #735 is merged with the updated IR structs, could you proceed with aligning this one to the new `WithBareClause`/`WithMatchClause` split? :)

prehnRA:
@bartblast Yep! I'll pick up here soon

lib/hologram/compiler/transformer.ex
```diff
@@ -920,4 +932,58 @@ defmodule Hologram.Compiler.Transformer do
...
+      do_and_else
+      |> Keyword.get(:do)
+      |> transform(context)
+      |> case do
+        %IR.Block{} = block ->
+          block
+
+        other ->
+          %IR.Block{expressions: [other]}
+      end
```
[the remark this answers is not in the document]
prehnRA: We can't do exactly this because technically checking whether something is a keyword list in a guard is not feasible-- can't check whether an arbitrary length list contains certain elements in a guard.

However, I think we need to do a runtime check and handle two cases in the execution, because I think the transformer will currently return the wrong structure for something like this:

```elixir
with [:a, :b, :c] do
end
```

That's kind of an extreme edge case, but I'm sure it'll prevent a strange error somewhere.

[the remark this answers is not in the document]
prehnRA: Or do I just check that the head of the list is `{:do, do_part}`? The only downside would be that it makes it order-dependent, and I'm not sure if we foresee the AST entries for do...else ever being in a different order.

bartblast:
Hey @prehnRA, is it ready for review?

prehnRA:
@coderabbitai review

prehnRA:
@bartblast Ready for review now I think

test/elixir/hologram/compiler/transformer_test.exs
```diff
@@ -6452,15 +6452,527 @@ defmodule Hologram.Compiler.TransformerTest do
...
+                       }
+                     ]
+                   }
+                 }
+               ],
+               else_clauses: []
+             }
+    end
   end
 end
```
bartblast, line 7511: The current transformer tests follow a convention where each test case has two variants: `"case name (AST from source code)"` and `"case name (AST from BEAM file)"`. Could you add the BEAM file variants for the `with` tests as well? You'll probably need to create module fixtures for this. They're important since BEAM ASTs can differ from source ASTs (e.g. variable versions).

lib/hologram/compiler/transformer.ex
```diff
@@ -920,4 +930,64 @@ defmodule Hologram.Compiler.Transformer do
         %IR.Variable{name: name, version: version}
     end
   end
+
+  defp transform_with_clause(
```
bartblast, line 934: Small thing - private functions should be alphabetically ordered, so `transform_with_body` should come before `transform_with_clause` (b before c), etc.

lib/hologram/compiler/transformer.ex
```diff
@@ -920,4 +930,64 @@ defmodule Hologram.Compiler.Transformer do
...
+      {:->, _meta, [[match], body]} ->
+        %IR.Clause{
+          match: transform(match, %{context | pattern?: true}),
+          guards: [],
+          body: transform(body, context)
+        }
+    end)
+  end
+
+  defp transform_with_else_clauses([{:else, {:__block__, [], []}}], _context), do: []
```
bartblast: Could you move the `[{:else, {:__block__, [], []}}]` clause before the `[{:else, else_ast}]` one? That way the empty else is handled first and the `when is_list(else_ast)` guard can be removed since the tuple case is already matched.

lib/hologram/compiler/transformer.ex
```diff
@@ -920,4 +930,64 @@ defmodule Hologram.Compiler.Transformer do
...
+    }
+  end
+
+  defp transform_with_clause(clause, context) do
+    %IR.WithBareClause{
+      expression: transform(clause, context)
+    }
+  end
+
+  defp transform_with_body(do_part, context) do
```
bartblast: The `transform_with_body/2` wrapping of single expressions in `%IR.Block{}` should be handled at the normalizer level instead. The normalizer already does this for `def`/`defp`, `defmodule`, `for`, `try`, and `->` clauses - could you add a `normalize/1` clause for `:with` that wraps the `do` body (and else clause bodies) via `normalize_block/1`? That way `transform_with_body/2` can be removed.

lib/hologram/compiler/transformer.ex
```diff
@@ -920,4 +930,64 @@ defmodule Hologram.Compiler.Transformer do
...
+      other ->
+        # This can be returned from transform if the body has only one expression.
+        # For consistency and better type-checking, we want to wrap it in a block
+        %IR.Block{expressions: [other]}
+    end
+  end
+
+  defp transform_with_else_clauses([], _context), do: []
+
+  defp transform_with_else_clauses([{:else, else_ast}], context) when is_list(else_ast) do
```
bartblast: I see that `transform_with_else_clauses` handles `{:when, ...}` in else clauses, but there doesn't seem to be a test for it.

test/elixir/hologram/compiler/transformer_test.exs
```diff
@@ -6452,15 +6452,527 @@ defmodule Hologram.Compiler.TransformerTest do
...
-  test "with" do
-    ast =
-      ast("""
-      with true <- true do
-        :ok
-      end
-      """)
-
-    assert transform(ast, %Context{}) == %Hologram.Compiler.IR.With{}
+  describe "with" do
```
bartblast, line 6473: The current transformer tests follow a convention where each test case has two variants: `"case name (AST from source code)"` and `"case name (AST from BEAM file)"`. Could you add the BEAM file variants for the `with` tests as well? They're important since BEAM ASTs can differ from source ASTs (e.g. variable versions).

bartblast, line 6473: Also, the existing transformer tests follow a convention of testing empty/single/multiple variants for each aspect (e.g. "single clause" / "multiple clauses", "single expression body" / "multiple expressions body"). Some of these combinations are missing here - for example, there's a test for a clause with guards but not for multiple clauses with guards, and bare clauses don't have single/multiple variants either. It's important to cover these combinations because the AST can be surprising sometimes (e.g. a single element might be represented differently than a list of elements).

test/elixir/hologram/compiler/transformer_test.exs
```diff
@@ -6452,15 +6452,527 @@ defmodule Hologram.Compiler.TransformerTest do
...
+        """)
+
+      assert transform(ast, %Context{}) == %IR.With{
+               clauses: [],
+               else_clauses: [],
+               body: %IR.Block{expressions: []}
+             }
+    end
+
+    test "multiple clauses and else clauses" do
```
bartblast: This test seems redundant if we have dedicated tests for single/multiple clauses and single/multiple else clauses individually. Each test ideally focuses on one aspect - that makes it easier to pinpoint what broke when a test fails.

test/elixir/hologram/compiler/transformer_test.exs
```diff
@@ -6452,15 +6452,527 @@ defmodule Hologram.Compiler.TransformerTest do
...
+             }
+
+      assert arrow_clause == %IR.WithMatchClause{
+               match: %IR.Variable{name: :y},
+               guards: [],
+               expression: %IR.Variable{name: :x}
+             }
+    end
+
+    test "with plain expression (function call)" do
```
bartblast: This one seems redundant.

test/elixir/hologram/compiler/transformer_test.exs
```diff
@@ -6452,15 +6452,527 @@ defmodule Hologram.Compiler.TransformerTest do
...
+                         name: :other
+                       }
+                     ]
+                   }
+                 }
+               ]
+             }
+    end
+
+    test "pin match" do
```
bartblast: The "pin match" and "pin in else clause" tests are actually testing that match positions are transformed in pattern mode - could you rename them to make that intent clearer?

test/elixir/hologram/compiler/transformer_test.exs
```diff
@@ -6452,15 +6452,527 @@ defmodule Hologram.Compiler.TransformerTest do
...
+                   match: %IR.Variable{name: :x},
+                   guards: [],
+                   expression: %IR.Variable{name: :y}
+                 }
+               ],
+               else_clauses: []
+             }
+    end
+
+    test "handles lists as base expressions" do
```
bartblast: This test also seems redundant - list transformation is handled by the general `transform/2`, so if bare clauses are already tested there's no need to test every possible expression type within them.

bartblast (changes requested):
Thanks! A few things I noticed:

bartblast:
Hi @prehnRA,                           

Checking in on the with expression work. I'm planning the next Hologram release and would love to include it. Any chance you'll have time to finalize this PR (and the rest of the series) soon?                                     

I can prioritize helping you get them across the finish line in whatever way is most useful. Just let me know.                                                                                        

Thanks!

prehnRA:
@bartblast I did a bit more on it today, but I still need to:

- Add tests for the BEAM file AST variants
- Add test cases for guards in else clauses
- Add some other missing cases (multiple clauses with guards, multiple bare clauses, etc).
- Add some normalizer test cases now that I've pushed some of the logic into the normalizer

I should be able to put in more work on it tomorrow, but I'm not sure if I'll get all of that finished tomorrow. Not sure what you want to do about that with respect to the release.

bartblast:
Hey @prehnRA, thanks for pushing this! I've got a conference on April 22 that will take me away from Hologram dev for a few days, but I'd still like to cut a release around that time according to the dev plan, with the `with` support included. I want to have a few days of runway before the conf to prepare, so time is a bit tight. How about this: do as much as you can across the PRs (transformer, encoder, interpreter) through Wednesday, and then I'll take over whatever's left. It doesn't have to be perfect - I can polish up the code and tests and refactor later. For now we just need to make sure all cases are covered and that it works correctly.

prehnRA:
@coderabbitai review

prehnRA:
@bartblast This should be ready for re-review now.

bartblast:
Awesome, thank you @prehnRA! 💜

PR #753, opened 2026-03-19
Fix :erlang.list_to_binary/1 handling of nested improper-list iodata

### LINKED ISSUE #751: Fix :erlang.list_to_binary/1 for nested improper-list iodata
First of two issues to adress #742.

The recursive collector in `:erlang.list_to_binary/1` already handled improper tails when the tail was a binary, but it rejects the case where the improper tail is another list. 

The date formatting path can produce nested cons-built iodata in exactly that shape, so valid iodata like "2022-01-01" fails with `ArgumentError: not an iolist term`.

mward-sudo:
Fixes #751

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `list_to_binary/1` to properly handle deeply nested improper-list iodata structures containing nested lists and mixed data types. The function now correctly processes complex nested list compositions with binaries, integers, and other iodata components, enabling accurate conversion to binary format.

### DIFF
--- a/assets/js/erlang/erlang.mjs
+++ b/assets/js/erlang/erlang.mjs
@@ -1615,6 +1615,8 @@ const Erlang = {
 
         if (Type.isBinary(tail)) {
           chunks.push(tail);
+        } else if (Type.isList(tail)) {
+          collect(tail);
         } else {
           Interpreter.raiseArgumentError(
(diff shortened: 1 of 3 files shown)


bartblast:
Thank you @mward-sudo! 💜

PS: I changed the base to master since master is kept clean for fast patch releases.

PR #754, opened 2026-03-19
Delegate :erlang.iolist_to_binary/1 to :erlang.list_to_binary/1 and align iodata errors

### LINKED ISSUE #752: Fix :erlang.iolist_to_binary/1 semantics and error handling
Second of two issues to address #742.

`:erlang.iolist_to_binary/1` currently does its own conversion by calling `:lists.flatten/1` and then turns the flattened terms into bytes/binaries. That is the wrong abstraction for OTP semantics.

There are problems with this approach:

1. It treats any `bitstring` as a passthrough value.
In the current code, `Type.isBitstring(ioListOrBinary)` returns early. But `:erlang.iolist_to_binary/1` should only accept binaries directly, not arbitrary non-byte-aligned `bitstrings`.

2. It relies on `lists.flatten/1`. That loses the actual `iolist` validation semantics.

mward-sudo:
Fixes #752

## Dependencies

- [x] #753 

## Fix

The fix identified is:

1. Change `:erlang.iolist_to_binary/1` to accept only binaries as direct passthrough.
2. For list input, delegate to `erlang.list_to_binary/1` instead of reimplementing conversion.
3. For obvious non-list, non-binary input, raise ArgumentError with `not an iodata term`
4. If delegated `erlang:list_to_binary/1` raises the specific boxed `ArgumentError` with `not an iolist term` remap that one error to `not an iodata term`.

Why that fix is correct:

- `:erlang.list_to_binary/1` already owns the real recursive iolist validation/conversion logic.
- `:erlang.iolist_to_binary/1` should differ mainly in accepted direct input and error wording.
- Delegating removes duplicated logic and keeps both functions aligned while still preserving their distinct OTP contracts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter input validation for binary conversion: invalid inputs now raise a consistent ArgumentError.
  * Preserves already-binary inputs and improves handling of nested/improper list shapes, including nested tails.

* **Tests**
  * Added comprehensive tests covering identity behavior, nested/improper list conversions, validation failures, and consistent error remapping.

### DIFF
--- a/assets/js/erlang/erlang.mjs
+++ b/assets/js/erlang/erlang.mjs
@@ -1376,35 +1376,38 @@ const Erlang = {
   // TODO: test
   // Start iolist_to_binary/1
   "iolist_to_binary/1": (ioListOrBinary) => {
-    // TODO: validate arg
-
-    if (Type.isBitstring(ioListOrBinary)) {
+    if (Type.isBinary(ioListOrBinary)) {
       return ioListOrBinary;
     }
 
-    const chunks = Erlang_Lists["flatten/1"](ioListOrBinary).data.map(
-      (term) => {
-        // TODO: validate list item (binary or integer allowed)
+    if (!Type.isList(ioListOrBinary)) {
+      Interpreter.raiseArgumentError(
+        Interpreter.buildArgumentErrorMsg(1, "not an iodata term"),
+      );
+    }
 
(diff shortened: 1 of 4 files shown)


bartblast:
Thank you @mward-sudo! 💜

PR #864, opened 2026-06-11
Support umbrella projects

### LINKED ISSUE #875: Umbrella applications
Hologram currently assumes a single-app project layout. Umbrellas break in three places:

- **Compile time**: asset building fails in child apps because paths are resolved against the current working directory instead of the umbrella's shared `deps/`.
- **Live reload**: `mix holo` crashes at the umbrella root, and reload-triggered recompiles write bundles to a different location than assets are served from, so pages 404 on their JS after the first reload.
- **Runtime**: the project's OTP application cannot be identified at the umbrella root, breaking endpoint resolution and asset serving. Identification also breaks in releases containing more than one app that depends on Hologram (e.g. a pages-library app alongside the endpoint app).

## Scope

- `mix compile` works from the umbrella root and from child apps
- `mix holo` from the umbrella root works end to end: asset serving, live reload watching source dirs across all child apps, recompiled bundles served immediately
- Pages can be defined in multiple child apps and are served through the one endpoint app (pages-library apps take the Hologram dep but no compiler entry), in dev and in releases
- npm packages for JS interop live at `<umbrella root>/assets`
- Tooling works from the umbrella root (`mix holo.routes`, `mix holo.gen.ai_rules`)
- End-to-end coverage via a dedicated umbrella test app with its own CI job

## Limitations

- One Hologram endpoint app per running BEAM instance - a second endpoint app started in the same instance raises a descriptive error. Multi-instance support is not planned: the use case has workarounds (pages routed under a path prefix such as `/admin` with a dedicated layout, or separate releases per app).
- `mix holo` from inside the endpoint app dir works but watches only that app's own sources - run it from the umbrella root.
- Ponchos and custom `:deps_path` locations are untested (ponchos behave like single-app projects).

m1ome:
Closes #875 

There is a problem with Hologram when you trying to use it in a Umbrella application.
Problem starts when you try to run out `mix holo` on a Umbrella app, it just crashes on live reloading.

If you go down deep into particular folder you will have a problem with building assets.
This PR fixes problem, but leaving limitation there is can be possible only one Hologram used in Umbrella application. (And it should also be a BC).

Which also IMO should be fixed/implemented in future, in e.g. (you can have Hologram for admin and Hologram for web-facing user-application). But i don't see how to quickly implement it without BC. 

## Summary by CodeRabbit

* **Improvements**
  * Live reload now monitors source directories across all umbrella child apps and single-app projects.
  * Asset and workspace path resolution improved for multi-app workspaces.
  * Default AI rules source path now points to the packaged dependency location.

* **Tests**
  * Added umbrella fixtures and tests validating live-reload directories and new directory-helper utilities.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -11,6 +11,7 @@
           ~r"/deps/",
           ~r"/node_modules/",
           "test/features/",
+          "test/elixir/support/fixtures/live_reload/umbrella",
           "test/elixir/support/fixtures/compiler/normalizer/module_7.ex",
           "test/elixir/support/fixtures/compiler/normalizer/module_8.ex"
         ]
--- a/lib/hologram/compiler.ex
+++ b/lib/hologram/compiler.ex
@@ -303,13 +303,18 @@ defmodule Hologram.Compiler do
       "--target=es2021"
     ]
 
-    project_node_modules_path = Path.join([Reflection.root_dir(), "assets", "node_modules"])
+    # Include both the workspace root (umbrella root or single-app root) and
+    # the current project root (umbrella child app or single-app root) so that
+    # whichever layout the user has, NODE_PATH points at their node_modules.
+    # Non-existent directories on NODE_PATH are silently ignored by Node.
+    project_node_modules_paths =
+      [Reflection.workspace_root_dir(), Reflection.root_dir()]
+      |> Enum.uniq()
+      |> Enum.map(&Path.join([&1, "assets", "node_modules"]))
 
     node_path =
-      Enum.join(
-        [opts[:node_modules_path], project_node_modules_path],
-        PathUtils.env_path_separator()
-      )
+      [opts[:node_modules_path] | project_node_modules_paths]
+      |> Enum.join(PathUtils.env_path_separator())
 
     esbuild_opts = [
       env: [{"NODE_PATH", node_path}],
--- a/lib/hologram/compiler.ex
+++ b/lib/hologram/compiler.ex
@@ -312,9 +312,10 @@ defmodule Hologram.Compiler do
       |> Enum.uniq()
       |> Enum.map(&Path.join([&1, "assets", "node_modules"]))
 
-    node_path =
-      [opts[:node_modules_path] | project_node_modules_paths]
-      |> Enum.join(PathUtils.env_path_separator())
+    node_path = Enum.join(
+      [opts[:node_modules_path] | project_node_modules_paths],
+      PathUtils.env_path_separator()
+    )
 
     esbuild_opts = [
       env: [{"NODE_PATH", node_path}],
--- a/lib/hologram/compiler.ex
+++ b/lib/hologram/compiler.ex
@@ -312,10 +312,11 @@ defmodule Hologram.Compiler do
       |> Enum.uniq()
       |> Enum.map(&Path.join([&1, "assets", "node_modules"]))
 
-    node_path = Enum.join(
-      [opts[:node_modules_path] | project_node_modules_paths],
-      PathUtils.env_path_separator()
-    )
+    node_path =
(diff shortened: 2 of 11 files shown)


m1ome:
@bartblast any thoughts on this?

bartblast:
Thanks @m1ome, the direction looks good. There are a few things to tweak, and I also need to think through the feature-testing side since this will probably need a separate end-to-end umbrella app to test against.

I've opened #875 to track this. Could you drop a comment there so I can assign it to you? Going forward I've stopped taking PRs without an issue first, since hashing out the right approach up front is by far the most productive route for me.

I'm heads-down until v0.10 ships and won't have bandwidth until the second half of the next week, but I'll get back to this then. Appreciate you digging into this!

bartblast:
@m1ome sorry for the long silence here. I've been swamped for months and this deserved a reply much sooner. Thanks for digging into it and sending a fix - you were the first to report that `mix holo` crashes on an umbrella, and that's what got this moving.

I've added umbrella support in #983, which is on `dev` and ships in v0.11, so I'm closing this in favour of that one. Wanted to explain why rather than just close it.

You were right about live reload, and #983 ends up with the same `watched_dirs/0` as yours: branch on `Mix.Project.apps_paths/0`, union every child app's `:elixirc_paths` through `Mix.Project.in_project/3`.

The problem is that's only the first failure. Three more show up once reload actually runs:

* Mix reads `:listeners` from the project it runs as, so declaring `Phoenix.CodeReloader` in the endpoint app leaves it unregistered at the umbrella root.
* Per-app `compile.lock` manifests read as permanently stale.
* The reloader compiles with `--purge-consolidation-path-if-stale`, which can leave a module loaded from a consolidated beam that's no longer on disk. `:code.which/1` then returns a dead path, our compiler dies with `:enoent`, and the crash leaves a stale lock behind, so every later request 500s until you restart.

With just the `watched_dirs/0` fix you get one good reload and then it wedges.

Two more outside reload. `Reflection.otp_app/0` is `Mix.Project.config()[:app]`, which is nil at an umbrella root, and its release fallback raises as soon as two apps depend on Hologram. And `static_dir` / `root_priv_dir/0` resolve against cwd while the endpoint serves `apps/<endpoint_app>/priv/static`, so pages 404 on their JS.

That's really why I went a different route. Almost everything that breaks under umbrellas is integration level (cwd, path resolution, boot order, reload) and unit tests over `watched_dirs/0` can't see any of it. So #983 adds a real umbrella feature tests suite under `test/umbrella` - three child apps (endpoint, a plain lib, and a pages-only library app) driven by Wallaby in a browser, covering rendering a page defined in a sibling app, running an action that calls a sibling app's code, npm imports, plus `mix holo.routes`, `mix holo.gen.agents_md` and live reload watch dirs. It runs as its own CI job. That suite is what found every item above, and I don't think this can be maintained without it.

On running multiple Hologram apps in one umbrella: I don't plan to support it, one is enough. #983 raises a descriptive error if a second endpoint app starts in the same instance, and the workarounds are fine (path prefix like `/admin` with its own layout, or separate releases). Umbrellas are pretty niche these days and the Elixir docs are cautious about them too, in a section called ["Don't drink the kool aid"](https://elixir.hexdocs.pm/main/dependencies-and-umbrella-projects.html#don-t-drink-the-kool-aid):

> If you find yourself in a position where you want to use different configurations in each application for the same dependency or use different dependency versions, then it is likely your codebase has grown beyond what umbrellas can provide.

So the goal I set for #983 is narrower than multi-app support: just make it possible for existing umbrella codebases to use Hologram at all. If that turns out to be wrong and people actually want more, I'll revisit.

Could you try `dev` on your real setup and let me know if it works? Yours is the case that started this and I'd rather hear about gaps before v0.11 than after.

PR #887, opened 2026-06-22
Add text selection primitives

olivermt:
Adds primitives for reading and restoring text selections, including offsets and direction from relevant input events.

This is needed for building rich text editors and other text-editing experiences that need precise cursor and selection state.

Tests cover the new JavaScript event payload handling, selection registry behavior, rendering integration, and Elixir deserialization/helpers.

## Summary by CodeRabbit

* **New Features**
  - Added DOM effects system enabling components to restore selections, focus elements, and apply text ranges after rendering
  - Added support for `beforeinput` and `selectionchange` event bindings for enhanced input handling

* **Tests**
  - Added comprehensive test coverage for DOM effects functionality and new event handlers

### DIFF
--- a/assets/js/component_registry.mjs
+++ b/assets/js/component_registry.mjs
@@ -9,6 +9,16 @@ export default class ComponentRegistry {
     ComponentRegistry.entries = Type.map();
   }
 
+  // Optimized (mutates dom_effects field in-place)
+  static clearDomEffects(cid) {
+    const entry = ComponentRegistry.entries.data[Type.encodeMapKey(cid)][1];
+    const componentStruct = entry.data["atom(struct)"][1];
+    componentStruct.data["atom(dom_effects)"] = [
+      Type.atom("dom_effects"),
+      Type.list(),
+    ];
+  }
+
   // Optimized (mutates next_action field in-place)
   static clearNextAction(cid) {
     const entry = ComponentRegistry.entries.data[Type.encodeMapKey(cid)][1];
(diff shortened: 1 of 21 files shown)


bartblast:
Hey Oliver, thanks for jumping on this! I think it got away from us a bit though - we'd talked about keeping it small, and this ended up as three features in one big PR (`$selection_change`, `$before_input`, and the DOM effects system).

So I'm going to close it, but let me lay out where my head's at on each.

The DOM effects part is genuinely something I want in Hologram eventually. The thing is, the approach here is imperative, and I'd want it declarative instead - something like `<input $selection={@range}>` with selection as controlled state. The events themselves would also need a few changes first. The main one: any payload field that's a DOM node has to be turned into a native Elixir term - probably a reference - that we map back to the node client-side. We can't address nodes by child-index paths the way this does, since those paths aren't stable across a re-render, so once the DOM changes they end up pointing at the wrong node.

And size-wise, it's already huge - with the feature tests it really needs, it'd be roughly double. Big PRs like this just aren't workable for me to review and own. What works far better is breaking things into granular feature-request issues so we can hash out the design first. Most of the work is the speccing anyway, and once that's settled it's much easier for me to just implement it myself.

I've already split the event side into two issues so we can work through them properly:
- #889
- #890

(Drop any thoughts or ideas in there.)

To be clear, this isn't about your work at all, it's about what's sustainable for me to own. And if anything in here is a blocker, something you need that Hologram can't do today, give me a shout and I'll gladly bump it up the priority list.

PR #888, opened 2026-06-22
Try to avoid Ash reflection pulling in whole resource graph

olivermt:
Draft for discussion.

In Evo Store, page bundles were pulling in large parts of the application because Ash resources are also Ecto schemas. When a server init/3 path touched a resource module, Hologram added the Ecto reflection MFAs: __changeset__/0, __schema__/1, and __schema__/2. The generated Ash/Ecto schema metadata then seemed to pull in the broader resource graph through relationship metadata.

We are not yet sure whether this is due to circular references, association metadata, or something specific to Ash reflection. It also has not been a problem in other Hologram systems so far.

This change stops page bundle MFA selection from adding reflection MFAs that are reachable only from server init/3. It keeps the existing Ecto reflection fallback when an Ecto schema module is reachable from client-side page/component code; the tests cover both cases.

Submitting this as a draft because there was an existing test asserting the opposite behavior, so the goal is to start a discussion on the right semantics and implementation path before treating this as the final fix.

## Summary by CodeRabbit

* **Refactor**
  * Simplified the compiler's call graph analysis by removing automatic reflection function augmentation from server-side component initializers, focusing on directly reachable code from page entry points.

* **Tests**
  * Updated test coverage and fixtures to validate the streamlined analysis behavior.

### DIFF
--- a/lib/hologram/compiler/call_graph.ex
+++ b/lib/hologram/compiler/call_graph.ex
@@ -637,12 +637,11 @@ defmodule Hologram.Compiler.CallGraph do
   @spec list_page_mfas(t, module) :: [mfa]
   def list_page_mfas(call_graph, page_module) do
     entry_mfas = list_page_entry_mfas(page_module)
-    graph = get_graph(call_graph)
 
-    graph
+    call_graph
+    |> get_graph()
     |> sorted_reachable_mfas(entry_mfas)
     |> reject_hex_mfas()
-    |> add_reflection_mfas_reachable_from_server_inits(page_module, graph)
     |> Enum.uniq()
     |> Enum.sort()
   end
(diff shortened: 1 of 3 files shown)


bartblast:
Thanks for digging into this, Oliver. The main thing holding me back is that this removes the symptom rather than the cause. As you say yourself: "We are not yet sure whether this is due to circular references, association metadata, or something specific to Ash reflection." Until we know exactly where the over-inclusion comes from, I can't tell whether a change to MFA selection is correct or just hiding the problem somewhere else.

I also can't take this as-is because it would break a lot of apps today. `__schema__` and `__changeset__` are needed on the client for Ecto-based validations - client-side form validation being the obvious case - and Ecto schemas and changesets are easily reachable through templates and actions in normal apps, not just server `init/3`. Whatever we do here has to be a general solution.

I've opened a triage issue to track this: #891. The next step is a minimal reproduction, ideally a fork of https://github.com/bartblast/hologram_skeleton that triggers the blow-up with the smallest possible schema setup. With that I can pin down whether it's circular references, association metadata, or something Ash-specific, and design a fix for the actual root cause.

I'm going to close this for now, since the fix won't live in this form, but the discussion continues in #891. If you can put together that repro there it would be a huge help. Thanks again for surfacing this.

PR #897, opened 2026-06-23
Implement `prevent_default` event modifier

### LINKED ISSUE #895: `prevent_default` event modifier
Event bindings preventDefault per event type via the `isDefaultAllowed` flag, overridable per binding with `allow_default`. The override only works one way: there is no declarative way to preventDefault on events that allow default by design (mainly keyboard), so patterns like Enter-to-send without a newline (`$key_down.enter`) are not expressible today. Add `prevent_default` as the mirror of `allow_default`.

## Tasks

- Add a `prevent_default` modifier (mirror of `allow_default`): parse in `lib/hologram/template/event_modifiers.ex`, read and thread through `assets/js/renderer.mjs`, gate in `assets/js/hologram.mjs` as `forcePreventDefault || (!isDefaultAllowed && !allowDefault)`.
- Raise at compile time when `prevent_default` and `allow_default` are both on one binding (same as `debounce` + `throttle`).
- Tests: feature test mirroring `allow_default_page.ex` using `$key_down.enter.prevent_default`, plus parsing and conflict tests in `event_modifiers_test.exs`.
- Docs: the `prevent_default` modifier.

bartblast:
Closes #895

## Summary by CodeRabbit

- **New Features**
  - Added a `prevent_default` event modifier to explicitly force native default prevention, including safe composition with other event modifiers.
- **Bug Fixes**
  - Improved event dispatch robustness when event payloads don’t implement `preventDefault` or `stopPropagation`, while still executing the bound action.
- **Documentation**
  - Updated Events docs to standardize debounce/throttle and revised `allow_default`/`prevent_default` and stop-propagation notation.
- **Tests**
  - Added unit and end-to-end coverage for `prevent_default`, including validation that it can’t be combined with `allow_default`.


bartblast:
@coderabbitai full review

PR #965, opened 2026-07-25
Make debounce and throttle lifecycle-aware

### LINKED ISSUE #963: Lifecycle-aware debounce and throttle
A debounce modifier on an event binding should only affect when the dispatch happens, never what the program observes. Currently a pending debounced dispatch is unaware of what happens around it:

- Submitting a form reads component state that still lags the DOM, and the trailing dispatch later re-enters the pre-submit value into state
- When the bound element loses focus, the dispatch keeps waiting out the remaining window even though no further events can arrive
- When the user navigates to another page, the dispatch fires anyway and executes an action in a context that no longer applies

A throttled binding shares the last problem: a trailing dispatch held inside an open throttle window also fires after page navigation.

bartblast:
Closes #963

## Summary by CodeRabbit

* **New Features**
  * Multiple event bindings on the same element now act independently and execute in written order.
  * Debounced handlers now deterministically flush on focus loss and before form submission.
  * Pending debounced and throttled dispatches are canceled when navigating to another page.
* **Documentation**
  * Updated Events guidance for binding order, independent debounce/throttle behavior (including key filters), and lifecycle/navigation boundaries.
* **Tests**
  * Added/updated end-to-end and JS tests to verify flushing, cancellation, and multi-binding interactions.


bartblast:
@coderabbitai full review

bartblast:
@coderabbitai full review

PR #985, opened 2026-08-04
Preserve sibling identity across template blocks

### LINKED ISSUE #927: Sibling elements lose DOM identity when a block's output size changes
A false `{%if}` renders nothing — the node vanishes from the child list. Since child vnodes are keyless, snabbdom then matches siblings positionally: when a conditional toggles on, a later sibling can get paired with the new element ("same tag, same position") and be patched into it — its real DOM node gutted and rebuilt, losing focus, scroll, media state, and re-decoding images. Whether this happens is coincidental: snabbdom's two-ended walk rescues the sibling only if the list's other end is stable (in our app, a dev-only live-reload iframe at the end of <body> was enough to break the rescue — the push banner appearing rebuilt the entire workspace).

Other frameworks all guarantee the invariant "a conditional's state never changes  sibling identity": React keeps a fixed slot (null when off), Vue renders a  anchor for false v-if (verified: the sibling's DOM node survives the toggle), Svelte compiles direct node references so no re-matching exists.

**Fix**
Render an anchor comment vnode for a false {%if} (Vue's approach) — list arity stays constant, positional matching stays correct, no template syntax changes. Template key support would be the complementary tool for reorderable lists.

**Workaround**
Keep a stable always-rendered wrapper around conditional siblings (e.g. a display: contents slot).

bartblast:
Closes #927

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DOM updates for conditional, looped, nested, and sibling content blocks.
  * Preserved form input values, focus, and element identity when nearby content changes.
  * Fixed handling of empty or whitespace-only blocks without affecting adjacent elements.

* **Tests**
  * Added comprehensive coverage for block rendering, nested markers, fragments, and stateful inputs.


bartblast:
@coderabbitai pause

Issue #1140, opened 2026-08-26
An integer and a float stop being equal inside a list, tuple or map


bartblast:
`==` treats an integer and a float of the same value as equal. `===` doesn't:

```elixir
1 == 1.0     # true
1 === 1.0    # false
```

On the client that difference disappears the moment the numbers sit inside something:

| | server | client |
| --- | --- | --- |
| `[1] == [1.0]` | true | false |
| `{:ok, 2} == {:ok, 2.0}` | true | false |
| `%{a: %{b: -7}} == %{a: %{b: -7.0}}` | true | false |
| `[1] != [1.0]` | false | true |

The loose comparison only happens at the top. Once it steps into a list, a tuple, or a map's values it switches to the strict one and stays there, so every number below the first level is compared as though `===` had been written. Any pair of the same value is affected, in either order, however deep it sits.

Numbers are the whole of it. `==` and `===` only part ways over integers and floats, so everything else compares the same under both.

Map keys are right and should stay that way. `%{1 => :x} == %{1.0 => :x}` is false on both sides, which is what Erlang does.

Nothing raises, so the only sign is a comparison quietly answering the wrong way. Mixing integers and floats in the same field is easy to do without meaning to, when a value arrives whole from one place and decimal from another.

Found while working on #986. Different bug, opposite direction: #986 is two maps comparing equal when they are not.

## Related

- #986

PR #1142, opened 2026-08-26
Keep == loose inside a list, tuple and map

bartblast:
Closes #1140

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved equality comparisons for lists, maps, tuples, and nested collections.
  * Integer and floating-point values with the same numeric value are now compared consistently within collections.
  * Strict equality now correctly distinguishes collection types and contents.

* **Tests**
  * Added coverage for nested collections, map keys and values, tuples, and improper lists.
  * Added end-to-end checks for numeric equality across supported comparison operators.


bartblast:
@coderabbitai pause

Issue #1151, opened 2026-08-28
Entity construction through the entity type


bartblast:
Making an entity means naming the type and then handing that type to another module.

```elixir
Account
|> Entity.new(handle: "bart")
```

Two lines for one thing, and it is not the spelling an Elixir dev tries first. `Account.new(handle: "bart")` is.

A bare struct has the related problem. `attribute :public, :boolean, default: false` says what an account carries when nobody sets one, but `%Account{}.public` comes back nil. The declaration and the struct disagree, and only the constructor knows which is right.

Existing code, tests and fixtures should move to whatever spelling this settles on.

Depends on #1124, which puts entity construction in the browser.

PR #1156, opened 2026-08-28
Construct an entity through its own type

bartblast:
Closes #1151

## Summary by CodeRabbit

- **New Features**
  - Entity types now provide their own `new` constructors for creating records with maps, keywords, or no values.
  - Bare structs and constructed entities consistently include declared default attribute values.
  - Added support for reading entity defaults directly from a bare struct.

- **Bug Fixes**
  - Non-create writes now correctly represent unspecified attributes using their declared defaults.
  - Role grants clearly reject direct construction and must be managed through role-grant operations.

- **Documentation**
  - Clarified entity construction, default values, and write-data behavior.


bartblast:
@coderabbitai pause

PR #1235, opened 2026-09-17
Classify the bundled modules from the module info PLT

### LINKED ISSUE #1234: Every page entry file reads the beam of every module it reaches to classify it
Rendering a page entry file asks two questions about every module the page reaches, once per page:

- `render_module_function_defs/6` calls `Reflection.protocol?/1`, which is `elixir_module?/1` and `has_function?(module, :__protocol__, 1)`;
- `list_js_import_modules/2`, through `aggregate_js_imports/3`, calls `Reflection.has_function?(module, :__js_imports__, 0)`.

Since #1222, `has_function?/3` answers for a module that is not loaded by reading the exports chunk of its beam, without loading it. So each of these checks on an unloaded module is a file read, and the checks run once per module per page.

Until #1232 this was hidden: the module metadata registration loaded every module a page reaches (`Code.ensure_loaded?/1`), and once a module is loaded `has_function?/3` asks the VM, which costs nothing. #1232 computes the metadata from the module info PLT and no longer loads the modules, so the checks now read the beams on every page.

Both answers are facts of the module that do not depend on the page. The module info PLT (#1223) already records `protocol?` from the export table; a `js_imports?` flag (`{:__js_imports__, 0} in exports`) can be recorded the same way. The renderers can then take the module info PLT and read both flags, falling back to today's checks when no PLT is given. `encode_reachable_functions/4` asks `Reflection.protocol?/1` once per module per compile as well and can read the same flag.

## Measurements

On an app with 9,018 modules and 158 pages, fresh VM, dev environment, after #1232, rendering all 158 pages from a filled encode cache:

| | per compile |
|---|---|
| `Reflection.protocol?/1` calls | 60,414 |
| `Reflection.has_function?/3` calls | 181,242 (two per `protocol?/1`, one per JS import check) |
| `:beam_lib.chunks/2` calls | 179,124 |
| the 158 page builds, first pass in a fresh VM | 14.4 s |
| the 158 page builds, once every reached module is loaded | 0.7 s |

With #1232 alone, the page entry file phase on that app takes 17.6 s with nothing changed (18.8 s before it), and a no-change dev build takes 49.0 s (39.9 s before it, when the metadata registration loaded the modules).

## Not measured

The page phase with both checks answered from the module info PLT. What `encode_reachable_functions/4`'s once-per-module check costs on its own.

bartblast:
Closes #1234

## Summary by CodeRabbit

- **New Features**
  - Added reflection support for identifying modules that provide JavaScript imports.
  - Added module metadata to compiler analysis, enabling more consistent protocol and JavaScript import detection.

- **Improvements**
  - Updated page and runtime compilation workflows to use cached module information.
  - Refreshed compiler benchmark coverage and performance measurements, including reachable-function encoding and entry-file generation.

- **Tests**
  - Expanded coverage for module metadata, protocol detection, JavaScript imports, and compiler behavior with and without cached information.


bartblast:
@coderabbitai pause

